PR #12, opened 2024-04-15
[core] Enable the CI

mnajdova:
Things done in the PR: 
- disabled the browser tests, they were never enabled for Pigment CSS
- added config for building codesandbox packages


.codesandbox/ci.json
```diff
@@ -0,0 +1,23 @@
+{
```
mnajdova, line 1: I forgot to add this in the previous PR.

pnpm-lock.yaml
```diff
@@ -15921,6 +15921,7 @@ packages:
       '@babel/helper-module-imports': 7.24.3
       '@babel/helper-plugin-utils': 7.24.0
       '@babel/parser': 7.24.4
+      '@babel/runtime': 7.24.4
```
mnajdova, line 15924:
```suggestion
```

.circleci/config.yml
```diff
@@ -259,37 +259,37 @@ jobs:
...
-          name: Coverage
-          command: |
-            curl -Os https://uploader.codecov.io/latest/linux/codecov
-            chmod +x codecov
-            ./codecov -t ${CODECOV_TOKEN} -Z -F "$REACT_VERSION-browser"
-      - store_artifacts:
-          # hardcoded in karma-webpack
-          path: /tmp/_karma_webpack_
-          destination: artifact-file
+  # test_browser:
```
mnajdova, line 262: I disabled the browser tests, they were never enabled for the Pigment CSS package. I left the code so that adding them back in the future will be easier.

.circleci/config.yml
```diff
@@ -147,13 +147,13 @@ jobs:
...
           name: Tests fake browser
           command: pnpm build && pnpm test:coverage:ci
-      - run:
-          name: Check coverage generated
-          command: |
-            if ! [[ -s coverage/lcov.info ]]
-            then
-              exit 1
-            fi
+      # - run:
```
mnajdova, line 150: I had to disable this, otherwise it was failing. I assume because the coverage hasn't been created yet.

michaldudak: That's worth investigating. The previous command should have generated the coverage report.

brijeshb42: This can be made default now since there are only Pigment packages here. The main issue here was that pigment's tests needed the built files to be available before running the tests but the default testing mechanism in core repo did not allow a way to do that.

michaldudak (approved):

.circleci/config.yml
mnajdova: Got it, let me update it then

.circleci/config.yml
```diff
@@ -146,7 +146,7 @@ jobs:
       - install_js
       - run:
           name: Tests fake browser
-          command: pnpm build && pnpm test:coverage:ci
```
mnajdova, line 149: I've updated the script to include the building

PR #13, opened 2024-04-15
[pigment-css][react] Implement sx transform for system components

brijeshb42:


brijeshb42:
The runtime overhead here is quite minimal but if that bit of extra performance is important to you, we could also add an option (in Pigment config) to not use this wrapper if you are sure that you are only using Pigment's `styled` components and there is not mix an match between using `emotion` or other runtime css-in-js solution.

packages/pigment-css-react/tests/sx/fixtures/sxProps.output.js
```diff
@@ -1,43 +1,52 @@
-import { styled as _styled } from '@pigment-css/react';
+import { sx as _sx8, styled as _styled } from '@pigment-css/react';
+import { ForwardSx as _ForwardSx4 } from '@pigment-css/react/private-runtime';
+import { sx as _sx6 } from '@pigment-css/react';
+import { ForwardSx as _ForwardSx3 } from '@pigment-css/react/private-runtime';
+import { sx as _sx4 } from '@pigment-css/react';
+import { ForwardSx as _ForwardSx2 } from '@pigment-css/react/private-runtime';
+import { sx as _sx2 } from '@pigment-css/react';
+import { ForwardSx as _ForwardSx } from '@pigment-css/react/private-runtime';
```
mnajdova, line 8: Why do we have this many duplicated imports? Can you explain how does this work?

brijeshb42: This is the generated output. For each usage of any of `css/styled/sx`, their processors imports their own required imports as the context is specific for each and is not shared across invocations.
So even if there are two `styled()` calls within the same file, they'll import and use their own identifiers.

mnajdova: > This is the generated output.

Yeah, I know, I was just surprised at how it looked like. I wondered if there is not a way to import it once and reuse the same import.

mnajdova (approved):

packages/pigment-css-react/tests/sx/fixtures/sxProps.output.js
o-alexandrov: If you'd like to add only one import per file, you can probably consider to add state outside of the import statements logic, for example:
```ts
// enum to decrease memory usage in cases w/ a lot of files (number vs string)
enum ImportType {
	forwardSx,
	sx,
	styled
}
type state = {
	[fileId: string]: {
		[importStatement in ImportType]?: true
	}
}
const state: state = {}

const addImportStatement = (fileId: string, importType: ImportType) => {
	if (!state[fileId]) state[fileId] = {}
	if (!state[fileId][importType]) {
		state[fileId][importType] = true
		// ... add import statement
	}
}
```

brijeshb42: Even if we did this for `Forward` sx, the imports added for `sx` itself can't be updated to do this since the context that `wyw` passes to each processor does not included a reference to the current file's ast. So it'll still continue to be the same for `sx/css/styled` calls.

PR #15, opened 2024-04-16
[core] Remove paths associated with the @mui packages

mnajdova:


packages/pigment-css-react/tsconfig.json
```diff
@@ -7,12 +7,6 @@
     "composite": true,
     "noEmit": false,
     "resolveJsonModule": true,
-    "paths": {
-      "@mui/system": ["./packages/mui-system/src"],
-      "@mui/system/*": ["./packages/mui-system/src/*"],
-      "@mui/utils": ["./packages/mui-utils/src"],
```
mnajdova, line 13: @brijeshb42 I noticed we don't list @mui/utils as direct dependency of @pigment-css/react, should we do it?

brijeshb42: We don't use it directly. Only through system. So it shouldn't be required.

mnajdova: I saw an import here: https://github.com/mui/pigment-css/blob/master/packages/pigment-css-react/src/createUseThemeProps.js#L1, this is why I asked.

mnajdova: Cool, I will add it.

brijeshb42 (approved):

PR #19, opened 2024-04-17
[core] Add GitHub workflows

mnajdova:
I isolated the commits for easier review. The first one just copies the files, while the second one is making the changes. I've altered the CONTRIBUTING.md guide too.


.github/ISSUE_TEMPLATE/5.priority-support.yml
```diff
@@ -0,0 +1,39 @@
+name: 'Priority Support: SLA ⏰'
```
michaldudak, line 1: I wonder if this should be here (same with Base UI). Do we provide priority support for X subscribers?

.github/workflows/no-response.yml
```diff
@@ -0,0 +1,37 @@
...
+on:
+  issues:
+    types: [closed]
+  issue_comment:
+    types: [created]
+  schedule:
+    # These runs in our repos are spread evenly throughout the day to avoid hitting rate limits.
+    # If you change this schedule, consider changing the remaining repositories as well.
+    # Runs at 12 am, 12 pm
+    - cron: '0 0,12 * * *'
```
michaldudak, line 14: These run times should be different per repository so we avoid running into API call limits on Github.

.github/ISSUE_TEMPLATE/5.priority-support.yml
mnajdova: I was wondering the same, there were few other X related things that I copied over, cc @oliviertassinari?

.github/workflows/no-response.yml
```diff
@@ -10,8 +10,8 @@ on:
   schedule:
     # These runs in our repos are spread evenly throughout the day to avoid hitting rate limits.
     # If you change this schedule, consider changing the remaining repositories as well.
-    # Runs at 12 am, 12 pm
-    - cron: '0 0,12 * * *'
+    # Runs at 1 am, 1 pm
+    - cron: '0 0, 1 * * *'
```
michaldudak, line 14: That's the correct expression for 1am, 1pm (see https://crontab.guru/#0_1,13_*_*_*):

```suggestion
    - cron: '0 1,13 * * *'
```

michaldudak (approved):

.github/ISSUE_TEMPLATE/5.priority-support.yml
oliviertassinari: https://github.com/mui/material-ui/pull/37671

oliviertassinari: We sold support on Core for a few customers, I believe, a handful. From their perspective, they need as much Material UI as the advanced components that extends Material UI, they compare us with Kendo UI or Syncfusion that offers the same product coverage. 

On the implementation, I believe the plan is to change this issue template to be broader, allow anyone to validate their support key and see what they have access to.

CONTRIBUTING.md
```diff
@@ -0,0 +1,234 @@
...
+### Trying changes in the apps
+
+While the documentation is still under development, we have few apps in the repository where you can test the changes. Before running the apps, make sure that the packages are build by running
+
+```bash
+pnpm build
+```
+
+After this, you can run one of the applications located in the `/apps` directory, by following their `README.md` files.
+
```
oliviertassinari, line 110: How about we delete all of this and instead apply https://github.com/mui/base-ui/blob/master/CONTRIBUTING.md?plain=1? It doesn't feel like we have bandwidth to maintain a hybrid version of this file, well.

PR #22, opened 2024-04-19
[react][core] Add support for sx replacement in transformed jsx calls

brijeshb42:
This got left out in the last PR where it only supported transformation in code with JSX.
This PR adds support for sx transform in already transformed JSX element to `jsx()` calls.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,6 @@
     "@types/lodash": "^4.17.0",
     "@types/mocha": "^10.0.6",
     "@types/node": "^18.19.30",
-    "@types/prettier": "^3.0.0",
     "@types/react": "^18.2.74",
     "@types/yargs": "^17.0.32",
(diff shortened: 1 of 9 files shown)


mnajdova (approved):

mnajdova:
@brijeshb42 what do you think about adding labels for each @pigment-css packages and add tags related to the packages. For example, I would title this PR: "[pigment-css-react][bug] Add support for sx replacements in transformed jsx calls"

brijeshb42:
Makes sense regarding the tags. I'd maybe skip the `pigment` in the name since this repo is only for Pigment CSS, like `[react][bug]`. WDYT?

brijeshb42:
@siriwatknp Let me know if this fixes your issues. I'll merge once you confirm.

PR #37, opened 2024-04-26
Enable `sx` prop on any components

### LINKED ISSUE #35: [react] Support the sx prop on all components
Regardless of whether they are built with PIgment CSS's styled API or not

**Search keywords**:

siriwatknp:
closes #35 

- transform `sx` prop using runtime `{…sx()}` spread (see the test cases)
- Remove `ForwardSx`
- Remove `allowTransformSx` check (to transform any component)
- Remove sx related logic from runtime `styled` (I like this so much because `sx` is not independent of `styled`)
- Remove double class CSS specificity

Tested in next-app, it works with HTML tags and Material UI components.

---


apps/pigment-css-next-app/.eslintrc
```diff
@@ -3,6 +3,7 @@
   "rules": {
     "import/prefer-default-export": "off",
     "import/extensions": "off",
-    "import/no-unresolved": "off"
-  }
+    "import/no-unresolved": "off",
+    "react/no-unknown-property": ["error", { "ignore": ["sx"] }],
```
siriwatknp, line 7: Ignore `sx` prop on JSX html tag.

apps/pigment-css-next-app/src/app/page.tsx
```diff
@@ -1,7 +1,17 @@
...
 import { styled, css } from '@pigment-css/react';
+import Button from '@mui/material/Button';
 import styles from './page.module.css';
 
+declare global {
+  namespace React {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    interface HTMLAttributes<T> {
+      sx?: any;
+    }
```
siriwatknp, line 11: Just to test that we can modify the types of JSX elements.

packages/pigment-css-react/src/processors/sx.ts
```diff
@@ -156,10 +133,136 @@ export class SxProcessor extends BaseProcessor {
...
+          );
+        }
+        sxAttribute.remove();
+      }
+      tagPath.node.arguments.push(t.objectExpression(attributes));
+      return tagPath.node;
+    }, false);
+
+    // For non-JSX calls, replace the sx prop with runtime sx
+    this.replacer((_tagPath) => {
```
brijeshb42, line 214: Why multiple replacer calls? Shouldn't it happen only once?

packages/pigment-css-react/src/processors/sx.ts
```diff
@@ -156,10 +133,133 @@ export class SxProcessor extends BaseProcessor {
...
+     */
+    this.replacer((_tagPath) => {
+      const tagPath = _tagPath as NodePath<CallExpression>;
+      return t.callExpression(tagPath.get('callee').node, [result]);
+    }, false);
+
+    /**
+     * For JSX calls, replace the sx prop with runtime sx
+     */
+    this.replacer((_tagPath) => {
```
siriwatknp, line 154: Casting type like this will fail `this.replacer((_tagPath: NodePath<CallExpression>) => {` because Type 'Node' is not assignable to type 'CallExpression'.

packages/pigment-css-react/src/processors/sx.ts
```diff
@@ -156,10 +133,133 @@ export class SxProcessor extends BaseProcessor {
...
+      const tagPath = _tagPath as NodePath<CallExpression>;
+      const jsxElement = tagPath.findParent((p) =>
+        p.isJSXOpeningElement(),
+      ) as NodePath<JSXOpeningElement>;
```
brijeshb42, line 158: It'd be nice to have this logic in a separate util and use here. Will keep this file clean.

apps/pigment-css-next-app/src/app/page.tsx
mnajdova: Nice, we can use the `SxProps` tough, no? :) I would recommend adding a documentation about this in the README.md too, so we can test everything, including typings and make sure we haven't forgot something.

apps/pigment-css-next-app/src/app/page.tsx
```diff
@@ -93,8 +103,16 @@ export default function Home() {
...
+      <Description
+        sx={{
+          '&::before': {
+            content: '"🚀"',
+            display: 'inline-block',
+            border: '1px solid',
+          },
+        }}
+      >
+        <p sx={{ boxShadow: '0 0 4px 0 rgba(0 0 0 / 0.12)' }}>
```
mnajdova, line 115: I ❤️ this!

packages/pigment-css-react/src/processors/sx.ts
siriwatknp: I tried with one replacement but I encountered a weird transformation, it does not work with all the test cases especially conditional and logical expressions, so I split the replacement into 2 steps. First, the existing logic of build-time sx and then replace it with runtime sx spread.

I can combine the replacement for JSX and non-jsx calls.

brijeshb42:
I haven't tested this but how would conditional sx work here -> 

```jsx
<div sx={props.isPrimary ? {color: 'red', background: 'blue'} : {fontWeight: '400'}} />
```

siriwatknp:
> I haven't tested this but how would conditional sx work here ->
> 
> ```js
> <div sx={props.isPrimary ? {color: 'red', background: 'blue'} : {fontWeight: '400'}} />
> ```

```js
<div {…props.isPrimary ? sx(…) : sx(…) } />
```

The logic that you had implemented remains the same. What I added is the conversion of `sx` prop to spread `{…sx()}` call.

siriwatknp:
image

I have to add `@pigment-css/react` to the root package.json

apps/pigment-css-next-app/src/app/page.tsx
siriwatknp: It's for POC at this point. There is one issue to fix from Pigment side which is delegating `sx` logic to the theme. Currently, Pigment is using `styleFunctionSx` from MUI System which is impossible to customize.

packages/pigment-css-react/src/processors/sx.ts
siriwatknp: Done!

packages/pigment-css-react/src/processors/sx.ts
siriwatknp: Done.

packages/pigment-css-react/tests/sx/fixtures/sx-jsx.output.js
```diff
@@ -3,8 +3,15 @@ function App() {
   return /*#__PURE__*/ _jsx(
     'div',
     {
+      sx: {
```
brijeshb42: This `sx` prop is not transformed.

packages/pigment-css-react/tests/sx/fixtures/sx-jsx.output.js
```diff
@@ -16,32 +23,50 @@ function App() {
     this,
   );
 }
-function App2() {
+function App2(props) {
   return /*#__PURE__*/ _jsx('div', {
+    sx: {
```
brijeshb42: same. it should be removed if it is transformed.

packages/pigment-css-react/tests/sx/fixtures/sx-jsx.output.js
siriwatknp: Could you check it again?

brijeshb42 (approved):

PR #40, opened 2024-04-30
Fix styled inheritance

siriwatknp:
## Issue

A Pigment styled component (child) is replaced by `as` value provided to the parent:

```js
const Child = styled('span')({ classes: ['child'] });

const Parent = styled(Child)({ classes: ['parent'] });

<Parent as="div" />; // <div class="parent" />
```

This behavior is different from Emotion & Styled-components where styles of child are propagated even `as` is provided to the parent.

## Expected

```js
const Child = styled('span')({ classes: ['child'] });

const Parent = styled(Child)({ classes: ['parent'] });

<Parent as="div" />; // <div class="child parent" />
```

### DIFF
--- a/packages/pigment-css-react/src/styled.js
+++ b/packages/pigment-css-react/src/styled.js
@@ -1,5 +1,4 @@
 /* eslint-disable no-underscore-dangle */
-/* eslint-disable react/prop-types */
 import * as React from 'react';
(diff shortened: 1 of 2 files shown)


brijeshb42 (approved):
Thanks

PR #57, opened 2024-05-08
Decouple `sx` prop from system and add docs

siriwatknp:
I decided to change the intention of this PR so that we can ship incrementally.

From Pigment CSS users, `sx` prop gives a convenient way to style elements without writing any import. The shorthand should not be enabled by default (it's too magical).
If they want shorthand, they need to add some theme config (a new feature for Pigment CSS).

So this PR is about decoupling the shorthand from System so that `sx` prop does not come with shorthand by default.

Then consider shorthand as another feature that can be worked separately.

---


README.md
```diff
@@ -467,7 +468,70 @@ function App() {
...
+};
+
+<AnyComponent sx={{ fontSize: 12, color: 'red' }} />;
+```
+
+> 💡 **Gotcha**: You don't need to pass `sx` prop to underlying component or DOM because it will be replaced with `className` and `style` props.
+
+The value provided to `sx` prop can be one of the following:
+
+- a plain style object (recommended)
```
mnajdova, line 497: A why would be nice.

README.md
```diff
@@ -467,7 +468,70 @@ function App() {
...
+- a callback function that receives the [theme object](#theming) then return a plain style object:
+
+  ```js
+  <div sx={(theme) => ({ color: theme.colors.primary })} />
```
mnajdova, line 501: Maybe we can add example that is not possible with an object, so that people are aware of what's the different use-case.

README.md
```diff
@@ -467,7 +468,70 @@ function App() {
...
+  ```
+
+- an array of plain style objects or callback functions. This is useful for applying conditional styles based on other variables:
+
+  ```js
+  <div
```
mnajdova, line 507: Great example!

mnajdova (commented):
> The shorthand properties (currently, the shorthand works but it's tied to MUI System styleFunctionSx). There is a plan to move to theme.unstable_sx so that developer can configure their own which will be a different PR.

This is very important, I would even wait with adding documentation until we have this. The sx prop currently is tight to a theme structure, that people may not be aware of. If we don't have a documentation about this, it looks like magic.

README.md
brijeshb42, line 497: I'd also mention static values.

README.md
```diff
@@ -467,7 +468,70 @@ function App() {
...
+
+The `sx` prop works on any component, including HTML elements and 3rd-party custom components as long as it is JSX.
+
+### Usage
+
+```js
+<div sx={{ display: 'flex', flexDirection: 'column' }}>
+```
+
+For a React component, it must spread the `className` and `style` to the underlying DOM element, otherwise the styles won't be applied.
```
aarongarciah, line 483: nit: spread is just a way of passing props. I think "pass" (or a better word?) would work best here since some devs might think they're forced to spread props.

```suggestion
For a React component, it must pass the `className` and `style` props to the underlying DOM element, otherwise the styles won't be applied.
```

README.md
```diff
@@ -467,7 +468,70 @@ function App() {
...
+      { color: 'red' },
+      selected && { fontWeight: 'bold' },
+      disabled ? (theme) => ({ opacity: theme.state.disabledOpacity }) : { opacity: 1 },
+    ]}
+  />
+  ```
+
+### Shorthand properties
+
+> ⏳ **Coming soon**: We are working on the shorthand properties for `sx` prop.
```
aarongarciah, line 518: This part seems to take for granted that users know about what shorthand properties are. Maybe a longer explanation, an example, or a link to an existing resource of why shorthand properties are useful is helpful.

README.md
brijeshb42, line 483: This line doesn't seem applicable. We can do without it because users don't have to care about the spread.

README.md
```diff
@@ -467,7 +468,70 @@ function App() {
...
+
+```js
+const AnyComponent = (props) => {
+  return <div {...props} />;
+};
+
+<AnyComponent sx={{ fontSize: 12, color: 'red' }} />;
+```
+
+> 💡 **Gotcha**: You don't need to pass `sx` prop to underlying component or DOM because it will be replaced with `className` and `style` props.
```
brijeshb42, line 493: What does this mean ?

README.md
brijeshb42: We can link to docs on mui.com

README.md
siriwatknp:
```js
// components/Custom.tsx
const AnyComponent = () => {
  return <div>Hello</div>;
};

// App.tsx
<AnyComponent sx={{ fontSize: 12, color: 'red' }} />;
```

How would the above snippet work if the user does not pass props to the element?

README.md
siriwatknp: I am afraid that the user will create an intermediate component and pass the `sx` prop down, but I think it can be removed.

README.md
siriwatknp: Can you give me an example?

siriwatknp: Added

README.md
siriwatknp: What do you mean by "not possible with an object"? Can you give me an example?

README.md
siriwatknp: I decided to remove the "Shorthand properties" section. This PR can be independent of it.

README.md
```diff
@@ -691,6 +692,68 @@ declare module '@pigment-css/react/theme' {
...
+
+To use `sx` prop on HTML element, you need to augment the `HTMLAttributes` interface. Add the following code to a file that is included in your tsconfig.json:
+
+```ts
+declare global {
+  namespace React {
+    interface HTMLAttributes<T> {
+      sx?:
+        | React.CSSProperties
+        | ((theme: Theme) => React.CSSProperties)
```
brijeshb42, line 750: We should also add where to import this `Theme` from.

packages/pigment-css-react/tests/sx/fixtures/sxProps.output.css
```diff
@@ -23,11 +23,11 @@
   color: var(--stzaibv-0);
 }
 .sazg8ol {
-  margin-bottom: 8px;
+  mb: 1px;
```
brijeshb42, line 26: Is this expected ?

README.md
siriwatknp: Added as a placeholder to imply that it comes from the consumer.

brijeshb42 (approved):

packages/pigment-css-react/tests/sx/fixtures/sxProps.output.css
siriwatknp: Fixed.

PR #63, opened 2024-05-09
[core] Cleanup

brijeshb42:


mnajdova (approved):
The changes makes sense. I would like to stress attention that we do want to have some regressions test on Pigment CSS, I know we are testing the utils via unit tests, but eventually we should add regressions tests, as it's the only way to have a real use-case scenario being tested with the library. This is the main reason why I didn't drop them so far, but considering how different the project is, I think it makes sense to do them from scratch.

brijeshb42:
Yes. I agree with having regression tests and I also had a similar thought of having it from scratch. That's why I removed those. There is some more cleanup to be done related to unit testing so that we can get rid of babel deps. But that will be a separate PR.

PR #71, opened 2024-05-10
[core] Migrate test runner to vitest

brijeshb42:
Note that the class names for the generated css have changed because earlier, the root directory was considered from repo root, but now, it considers the package directory as the root. And classname generation depends on this path.


oliviertassinari:
What's the motivation for this change?

Janpot:
For historic reasons Toolpad ended up using `vitest` instead of `mocha` for its unit tests. Current Toolpad could benefit from some of the test infra of core and maybe we'll try again to switch, but I have to admit, it's a lot less painful compared to karma+babel+mocha. Can't explain the the motivation for this change, but I can share my experience with `vitest`:

pros:
* faster and parallel
* no need for babel, runs typescript/jsx code out-of-the-box
* actually handles module resolution ~correctly~ as expected
* built-in mocking (looking at you Datepickers for testing multiple versions of date-fns)
* no need for karma to run tests in the browser
* it just works

cons:
* can't reuse mui test utils
* no babel means potentially testing differently compiled code.

oliviertassinari:
How do we run the tests in a real browser with vitest? As I understand things:

- The reason we use mocha is so we can run tests in a browser.
- We can't use playwright because it's x10 slower than Karma
- We can't use jsdom because it's not accurate enough, at least for the data grid, I believe @romgrk experience was too limited.

Janpot:
> How do we run the tests in a real browser with vitest?

Yep, they're adding a [browser mode](https://vitest.dev/guide/browser). Granted, it's not stable yet, but I made it work today in Toolpad in relatively short time. It was that or move to mocha+karma. I gave it a quick shot, and it's quite delightful, still not supporting the full functionality of the node.js runner, but certainly seems to be usable for us.

JCQuintas:
> We should transpile the source to be close to production, but I guess this won't be an issue.

Would moving to vite instead of babel be a huge issue? We have some complex babel transpiling in X that I'm not sure would be portable to vite.

Janpot:
> We have some complex babel transpiling in X that I'm not sure would be portable to vite.

Which transformations do you think aren't portable? The one I'm aware of is the macro in MUI core for error messages. I don't think X is even making use of that? In any case, I don't think there is anything preventing us from using a babel transform in vite if necessary.

Janpot:
> But I would guess on X the date adapters import replacement could be an issue.

😄 Actually mentioning this use case in my [comment](https://github.com/mui/pigment-css/pull/71#issuecomment-2150389634) above.

> * built-in mocking (looking at you Datepickers for testing multiple versions of date-fns)

I expect that we can build that on top of vitest [module mocking](https://vitest.dev/guide/mocking#modules).

PR #94, opened 2024-05-20
[feature] Implement useTheme processor

brijeshb42:
to replace the useTheme function calls with an import from Pigment package.
Also re-export `RtlProvider` from `@mui/system`.

Fixes https://github.com/mui/material-ui/issues/42260

There will be a follow-up PR to implement the runtime part from a generated source at build-time.

### DIFF
--- a/packages/pigment-css-react/src/processors/useTheme.ts
+++ b/packages/pigment-css-react/src/processors/useTheme.ts
@@ -1,0 +1,64 @@
+import type { Expression } from '@babel/types';
+import { validateParams, type Params, type TailProcessorParams } from '@wyw-in-js/processor-utils';
+import { type Replacements, type Rules } from '@wyw-in-js/shared';
+import BaseProcessor from './base-processor';
+
+export type Primitive = string | number | boolean | null | undefined;
+
+export type TemplateCallback = (params: Record<string, unknown> | undefined) => string | number;
+
+export class UseThemeProcessor extends BaseProcessor {
+  constructor(params: Params, ...args: TailProcessorParams) {
+    super([params[0]], ...args);
+    validateParams(params, ['callee', ['call']], `Invalid use of ${this.tagSource.imported} tag.`);
+  }
+
+  build() {
+    const cssText = '/* */';
+    const rules: Rules = {
+      [this.asSelector]: {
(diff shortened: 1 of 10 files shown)


packages/pigment-css-react/tests/useTheme/fixtures/useTheme.output.js
```diff
@@ -0,0 +1,117 @@
+import { useTheme as _useTheme } from '@pigment-css/react';
```
siriwatknp, line 1: Where is the runtime `useTheme`?

packages/pigment-css-react/src/processors/useTheme.ts
```diff
@@ -0,0 +1,64 @@
+import type { Expression } from '@babel/types';
+import { validateParams, type Params, type TailProcessorParams } from '@wyw-in-js/processor-utils';
+import { type Replacements, type Rules } from '@wyw-in-js/shared';
+import BaseProcessor from './base-processor';
+
+export type Primitive = string | number | boolean | null | undefined;
+
+export type TemplateCallback = (params: Record<string, unknown> | undefined) => string | number;
+
+export class UseThemeProcessor extends BaseProcessor {
```
siriwatknp, line 10: I thought that we were going with option B from https://www.notion.so/mui-org/material-ui-pigment-non-processor-replacement-0c7e4b150c90499b93cd8fc17c31a009?pvs=4

brijeshb42: I realised that tagProcessor handles all the different kinds of imports that we can have, ie,

```js
import * as lib from 'lib';
import {useTheme} from 'lib';
require('lib')
interopRequire(require('lib'));
```

Doing it ourselves might miss something especially for library code where most of the code is transpiled.

packages/pigment-css-react/tests/useTheme/fixtures/useTheme.output.js
brijeshb42: Here's the main PR now - https://github.com/mui/pigment-css/pull/105

See the description.

brijeshb42:
Closing in favor or #105

PR #105, opened 2024-05-27
[core] Implement useTheme processor and runtime theme

brijeshb42:
Here's how the replacement works

Before transformation:

```js
import { useTheme } from '../zero-styled';

function Component() {
  const theme = useTheme();
  // ...rest
}
```

After transformation:

```js
import _theme from '@pigment-css/react/theme';

function Component() {
  const theme = _theme;
  // ...rest
}
```

Fixes: https://github.com/mui/material-ui/issues/42260


packages/pigment-css-react/src/utils/generateThemeSource.ts
```diff
@@ -0,0 +1,24 @@
...
+  }
+  const hasBreakpoints = !!theme.unstable_args?.breakpoints;
+  const hasTransitions = !!theme.unstable_args?.transitions;
+  const breakpointsImport = hasBreakpoints
+    ? "import createBreakpoints from '@mui/system/createTheme/createBreakpoints';\n"
+    : '';
+  const transitionsImport = hasTransitions
+    ? `import createTransitions from '${process.env.PACKAGE_NAME}/createTransitions';\n`
+    : '';
+  const source = `${breakpointsImport}${transitionsImport}
```
siriwatknp, line 16: I think this should come from Material UI. Pigment should just call `theme.toRuntime()` and let Material UI decide what to add to the runtime.

In Material UI, we can export an adapter function that inserts `.toRuntime()` to the theme.

```js
// next.config.js
import { createPigmentConfig, extendTheme } from '@mui/material/styles';

const config = createPigmentConfig(extendTheme());

module.exports = withPigment(config);
```

siriwatknp: If you agree with my proposal, I can create a PR to add the `createPigmentConfig()`.

mnajdova: Agree with having this in Material UI side, and by default Pigment CSS can only add the CSS vars. What do you think?

siriwatknp: Yes, I think Pigment should call the specific method rather than owning the logic inside.

To let Pigment generate CSS variables from the theme:

> create a method `generateStyleSheets` to `theme`. At build time, Pigment will create CSS file from the result of the method.

To create a runtime theme:

> create a method `toRuntime` to `theme`. At build time, Pigment will use that result and export as JS (the result has to be a string)

To use the runtime theme:

> import theme from '@pigment-css/react/theme';

siriwatknp:
Here is my proposal, Material UI export a `stringifyTheme` function https://github.com/mui/material-ui/pull/42476 for zero-runtime libraries.

From Pigment side, I see 2 choices:

1. Pigment calls a method from the theme, e.g. `theme.toRuntime()`:

  ```js
   // next.config.js
  const { extendTheme, stringifyTheme } = require('@mui/material/styles');
  
  const theme = extendTheme();
  
  theme.toRuntime = () => stringifyTheme(theme);
  
  module.exports = withPigment({
    theme,
  })
  ```

2. Pigment config

  ```js
   // next.config.js
  const { extendTheme, stringifyTheme } = require('@mui/material/styles');
  
  const theme = extendTheme();
  
  module.exports = withPigment({
    theme,
    runtimeTheme: stringifyTheme(theme),
  })
  ```

I like this approach because regardless of the choice, it's not related to Material UI.

brijeshb42:
I like the first approach better, ie, `theme.toRuntime`. This way, users won't have to care about anything extra. We can make `toRuntime` method part of the `createTheme`/`extendTheme` from material-ui source code.
In future, this will anyways won't be part of the Pigment v1 release since the whole thing is very specific to Material UI. So we can move this over to the Material specific package.

I am naming the method to `toRuntimeSource` to be more explicit.

packages/pigment-css-react/src/processors/useTheme.ts
```diff
@@ -0,0 +1,61 @@
...
+    const themeIdentifier = t.addDefaultImport(`${process.env.PACKAGE_NAME as string}/theme`);
+    this.replacer(themeIdentifier, false);
+  }
+
+  get asSelector() {
+    return this.className;
+  }
+
+  get value(): Expression {
+    return this.astService.stringLiteral(this.className);
```
siriwatknp, line 59:
```suggestion
    return this.astService.nullLiteral();
```

`null` makes more sense.

packages/pigment-css-react/src/utils/extendTheme.ts
```diff
@@ -159,6 +160,10 @@ export function extendTheme<
       [this.getColorSchemeSelector(colorScheme)]: styles,
     };
   };
+  finalTheme.unstable_args = {
+    breakpoints: theme.breakpoints,
+    transitions: theme.transitions,
+  };
```
siriwatknp, line 166: Please remove this and the type above.

packages/pigment-css-react/tests/useTheme/fixtures/useTheme.input.js
```diff
@@ -0,0 +1,125 @@
...
+    onExit,
+    onExited,
+    onExiting,
+    style,
+    timeout = defaultTimeout,
+    // eslint-disable-next-line react/prop-types
+    TransitionComponent = Transition,
+    ...other
+  } = props;
+
```
siriwatknp, line 28: I think the test input should be simplified. A basic component that imports `useTheme()` should be enough.

siriwatknp (approved):
Preapproved, please check my suggestions before merging.

PR #111, opened 2024-05-28
[fix] Implement variant matching parity

DiegoAndai:
Follow up on https://github.com/mui/material-ui/pull/42358#pullrequestreview-2081441990

Merge props and `ownerState` the same way it's done in MUI System's `createStyled`

cc: @mnajdova

### DIFF
--- a/packages/pigment-css-react/src/styled.js
+++ b/packages/pigment-css-react/src/styled.js
@@ -3,12 +3,11 @@ import * as React from 'react';
 import clsx from 'clsx';
 import isPropValid from '@emotion/is-prop-valid';
 
-function getVariantClasses(componentProps, variants) {
-  const { ownerState = {} } = componentProps;
+function getVariantClasses({ ownerState = {}, ...componentProps }, variants) {
   const variantClasses = variants
     .filter(({ props: variantProps }) =>
       typeof variantProps === 'function'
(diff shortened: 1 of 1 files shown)


DiegoAndai:
**Before and after**

### Textfield demos with `0.0.10`:


https://github.com/mui/pigment-css/assets/16889233/c0c4fe61-4451-4a55-8128-c6a6ff6077cd

### Textfield demos with this PR's build:


https://github.com/mui/pigment-css/assets/16889233/58b05050-db8e-449c-b920-0c3bce9ed46a

-------


@brijeshb42 when running the demos with this PR's build, the following code on one of the demos errored:

```jsx
function RedBar() {
  return (
    <Box
      sx={{
        height: 20,
        backgroundColor: (theme) =>
          theme.palette.mode === 'light'
            ? 'rgba(255, 0, 0, 0.1)'
            : 'rgb(255 132 132 / 25%)',
      }}
    />
  );
}
```

With:

```
theme is not defined
```

This **didn't happen when running with `0.0.10`**. Why might this be? I would expect it to fail on both versions, specially as this change is not related to the `sx` prop or theme handling.

brijeshb42:
I'll have to check. There was a change by @siriwatknp where he change the callback argument from being `({theme}) =>` to `(theme) => `. This might be the reason.

brijeshb42 (approved):
Thanks.

PR #112, opened 2024-05-28
[react] Add `createExtendSxProp` for Material UI integration

siriwatknp:
⚠️ This PR is intended for internal use only. Documentation is not needed.


packages/pigment-css-react/src/createExtendSxProp.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Runtime function for creating `extendSxProp`.
+ *
+ * This is a placeholder for the runtime function for replacing the `extendSxProp` in Material UI
+ * to reduce bundle size only. DO NOT use it in your codebase.
+ */
+function extendSxProp(props) {
+  return props;
+}
+export default function createExtendSxProp() {
```
brijeshb42, line 10: Let's also mark it as `@internal` for later documentation

brijeshb42: Can you also add a fixture test for it ?

siriwatknp: Sure, I will.

mnajdova: To be safe, should we rename it to: `internal_createExtendSxProp`, maybe even in @mui/material we can keep this name.

siriwatknp: Done!

brijeshb42 (approved):

Issue #83, opened 2024-05-13
How can I custom a theme with NX+Next.js?


JenniferGoijman:
### Search keywords

NX, Next.js

### Related page

https://github.com/mui/pigment-css?tab=readme-ov-file#start-with-nextjs

### Kind of issue

Missing information

### Issue description

I have configured PigmentCSS in a NX+Next.js repo like this:

```
//@ts-check

// eslint-disable-next-line @typescript-eslint/no-var-requires
const { composePlugins, withNx } = require('@nx/next');
const { withPigment } = require('@pigment-css/nextjs-plugin');

/**
 * @type {import('@nx/next/plugins/with-nx').WithNxOptions}
 **/
const nextConfig = {
  nx: {
    // Set this to true if you would like to to use SVGR
    // See: https://github.com/gregberge/svgr
    svgr: false,
  },
};

const plugins = [
  // Add more Next.js plugins to this list if needed.
  withNx,
  withPigment
];

module.exports = composePlugins(...plugins)(nextConfig);
```

How can I custom a theme with NX+Next.js?

### Context

```
@pigment-css/react: 0.0.11
next: 14.0.4
react: 18.2.0
react-dom: 18.2.0
@nx/next: 18.0.2
@pigment-css/nextjs-plugin: 0.0.11
nx: 18.0.2
```

brijeshb42:
This requires a change in the api of `withPigment` to instead take in it's config from the `nextConfig` itself instead of taking it as a 2nd argument.
This will be a breaking change but makes sense if majority of other plugins are already following this approach.

omerman:
Getting this error on a setup with nx monorepo.
<img width="1085" alt="image" src="https://github.com/user-attachments/assets/f599c68b-6468-48d1-a35d-4920e4ccef56">

```js
//@ts-check

// eslint-disable-next-line @typescript-eslint/no-var-requires
const { createTheme } = require('@mui/material');
const { composePlugins, withNx } = require('@nx/next');
const { withPigment } = require('@pigment-css/nextjs-plugin');
require('../../core/src/next/utils/prepareEnv');

/**
 * @type {import('@nx/next/plugins/with-nx').WithNxOptions}
 **/
const nextConfig = {
  nx: {
    // Set this to true if you would like to use SVGR
    // See: https://github.com/gregberge/svgr
    svgr: false,
  },

  eslint: {
    ignoreDuringBuilds: false,
  },
[...]

PR #114, opened 2024-05-30
[nextjs] Update the withPigment API to accept config

brijeshb42:
through `pigment` key from the nextConfig object itself instead of as a 2nd argument to the call. This is a standard approach followed by other nextjs plugins

Fixes: #83


mnajdova (commented):
I am 👌 with this change, we just have to make sure we document it as a breaking change in the changelog

brijeshb42:
I'd say it's a deprecates the current API as it can still take the input as second argument. But I'll update the docs to note this.

siriwatknp:
I am dropping my thought here based on my user expectation and writing docs.

I would like all bundler plugin to use the same name, e.g. `createPigmentPlugin`:

**Next.js**

```js
import { createPigmentPlugin } from '@pigment-css/nextjs-plugin';

const pigmentPlugin = createPigmentPlugin(…pigment config);

export default pigmentPlugin(…next config)
```

**Vite**:

```js
import { createPigmentPlugin } from '@pigment-css/vite-plugin';

const pigmentPlugin = createPigmentPlugin(…pigment config);

export default defineConfig({
  plugins: [pigmentPlugin()]
})
```

With the above signature, writing docs will be a lot easier because we can refer to the setup once and later refer to only "pigment config" for other feature.

brijeshb42:
Seems like now Next.js warns users when it finds an unknown key in the config

```sh
 ⚠ Invalid next.config.mjs options detected: 
 ⚠     Unrecognized key(s) in object: 'pigment'
 ⚠ See more info here: https://nextjs.org/docs/messages/invalid-next-config
```

Now, I am not so sure that we should do this.

PR #118, opened 2024-05-31
[react] Create Stack component 

brijeshb42:
This uses atomic class generation utility to cover all the scenarios of rendering a Stack.


packages/pigment-css-react/src/baseAtomics.js
```diff
@@ -0,0 +1,43 @@
...
+      'flex-end',
+      'flex-start',
+      'self-end',
+      'self-start',
+      'start',
+      'baseline',
+      'normal',
+      'stretch',
+    ],
+    gap: ['--Stack-gap'],
```
brijeshb42, line 36: Do we need more properties to be added here?

packages/pigment-css-react/src/baseAtomics.js
```diff
@@ -0,0 +1,43 @@
+import { generateAtomics } from './generateAtomics';
+
+export const stackAtomics = generateAtomics(({ theme }) => ({
+  conditions: Object.keys(theme.breakpoints.values).reduce((acc, breakpoint) => {
+    acc[breakpoint] = `@media (min-width: ${theme.breakpoints.values[breakpoint]}${
+      theme.breakpoints.unit ?? 'px'
+    })`;
+    return acc;
+  }, {}),
+  defaultCondition: theme.breakpoints?.keys?.[0] ?? 'xs',
```
brijeshb42, line 10: Something to figure out. Right now, it's hardcoded if not present in theme.

brijeshb42:
Adding more commits with tests.

packages/pigment-css-react/src/base.ts
```diff
@@ -64,3 +65,10 @@ export interface PolymorphicComponent<SxProp, BaseProps extends BaseDefaultProps
     props: PolymorphicComponentProps<SxProp, BaseProps, AsTarget>,
   ): JSX.Element;
 }
+
+export interface BreakpointOverrides {}
+
+export type Breakpoint = OverridableStringUnion<
+  'xs' | 'sm' | 'md' | 'lg' | 'xl',
```
oliviertassinari, line 72: Seeing this makes me think of https://github.com/mui/material-ui/issues/26168

oliviertassinari:
We will need to be careful with the docs if we have `import { Stack } from "@pigment-css/react"`. People could perceive this as:

- lack of focus
- an API they are supposed to use and break false impressions of the library

mnajdova:
> We will need to be careful with how we document this if we have import { Stack } from "@pigment-css/react". People could perceive this as:
> 
> - lack of focus
> - an API they are supposed to use over the other API and create false first impressions about Pigment CSS

@brijeshb42 correct me if I am wrong, but we decided to create this in the Material UI package, no?

brijeshb42:
@mnajdova Creating `Stack` in material-ui means also adding Pigment CSS as a dependency. That's why I moved it to this package itself. I've discussed in the meeting as well if you've seen the video.

DiegoAndai:
> @brijeshb42 correct me if I am wrong, but we decided to create this in the Material UI package, no?

We did 

> Creating Stack in material-ui means also adding Pigment CSS as a dependency.

Why is that? Isn't it possible to migrate the layout components the same way the other components were migrated?

mnajdova:
> **_Marija_**: Agree for the components to be part of Material UI. Can it be peer dependency? Or maybe another question, if we are generating the CSS during the build and packaging it with the library would it still be required?
> **_Brijesh_**: If we are packaging it, then we get to choose if we want Pigment as a dependency of not. Right now, adding Pigment as a dependency adds a lot of packages that are required for the build but not part of the runtime. In this case, we can just have a @pigment-css/runtime package that would only have the runtime code and we can make them a dependency. Otherwise, we can bundle Pigment's runtime in materials package

This is the last conversation we had around this. I would be happy with any solution we pick right now, as long as the components come from Material UI. Why add technical debt when we know we are going to want this done anyway.

brijeshb42:
The current solution does not package things with Material UI. We have yet to start on the exploration for generating css before publishing.

DiegoAndai:
I reiterate my question 😅: Isn't it possible to migrate the layout components the same way the other components were migrated?

packages/pigment-css-react/src/base.ts
brijeshb42: We'll still need to continue supporting this unless we want breaking changes.
We could also introduce a way to alias breakpoints, ie, `xs -> mobile` etc.

packages/pigment-css-react/src/baseAtomics.js
```diff
@@ -0,0 +1,43 @@
...
+      'normal',
+      'stretch',
+    ],
+    gap: ['--Stack-gap'],
+  },
+  shorthands: {
+    direction: ['flexDirection'],
+    spacing: ['gap'],
+  },
+  multiplier: 8,
```
DiegoAndai, line 42: Did you figure out how to take this from the theme?

packages/pigment-css-react/src/baseAtomics.js
DiegoAndai: Should we figure this out before merging this PR?

brijeshb42: It already takes the value from theme. I made the first comment way back. It's not valid now.

packages/pigment-css-react/src/baseAtomics.js
brijeshb42: @siriwatknp What's your opinion here? `theme.spacing()` without arguments does not guarantee a numeric value.

siriwatknp: I think it should always be a string, then use CSS `calc()` as a value. To leverage the theme, use 

```js
multiplier: theme.vars.spacing, // it can be a string or an array.
```

packages/pigment-css-react/src/baseAtomics.js
```diff
@@ -0,0 +1,43 @@
+import { generateAtomics } from './generateAtomics';
+
+export const stackAtomics = generateAtomics(({ theme }) => ({
```
siriwatknp, line 3: I wonder why `stockAtomics` is not declared in `Stack.jsx`, is it intended to be used elsewhere?

siriwatknp (commented):
I am a bit confused. Is this a Stack component for Pigment CSS or for Material UI?

If it's for Material UI, based on the existing implementation `createStack`, only `direction` and `spacing` are the props because the rest of the system props are deprecated.

packages/pigment-css-react/src/baseAtomics.js
DiegoAndai: Having the atomics generator on the component file makes sense to me as well 👍🏼

DiegoAndai:
> I am a bit confused. Is this a Stack component for Pigment CSS or for Material UI?
> 
> If it's for Material UI, based on the existing implementation createStack, only direction and spacing are the props because the rest of the system props are deprecated.

This is the PigmentStack that Material UI will export. You're correct, following https://github.com/mui/material-ui/issues/42259, we can remove the `display`, `alignItems`, and `justifyContent` props. No need to have atomic classes for this as `sx` supports breakpoints. @brijeshb42 if you need to add an initial value to `display`, you can use the `styled` function, I [did this in the Grid component](https://github.com/brijeshb42/pigment-css/pull/1/files#diff-776f8e7258c2bdb171c06f1dae8695dfbec37ed85c54501a08b533166c48ceefR20-R25).

siriwatknp:
@DiegoAndai I pushed changes that simplify the logic in https://github.com/mui/pigment-css/pull/118/commits/ff1ef613ad7f0c64d14db1f31e64b0770afe4b7f:

- remove the `unitless` as I don't think we need it yet.
- multiplier should be a `string | undefined` from `theme.vars.spacing`
- remove `$$default` as I think `defaultCondition` suffices.
- for default condition, the variable name will not be suffixed with the breakpoint name. I think this is cleaner and less code.

apps/pigment-css-next-app/next.config.js
```diff
@@ -96,14 +96,40 @@ theme.getColorSchemeSelector = (colorScheme) => {
...
+
 /**
  * @type {PigmentOptions}
  */
 const pigmentOptions = {
   theme,
   transformLibraries: ['local-ui-lib', '@mui/material'],
   sourceMap: true,
-  displayName: true,
+  overrideContext: (context) => {
```
siriwatknp, line 113: Prevent an error `Cannot redefine property toString`

apps/pigment-css-next-app/src/app/page.tsx
```diff
@@ -112,6 +113,11 @@ export default function Home() {
           },
         }}
       >
+        <Stack spacing={{ xs: 2, md: 3 }}>
+          <div>Item1</div>
+          <div>Item1</div>
+          <div>Item1</div>
+        </Stack>
```
siriwatknp, line 120: Here is the result:  

image

packages/pigment-css-react/src/generateAtomics.js
```diff
@@ -11,37 +11,73 @@ export function generateAtomics() {
...
-      Object.keys(values).forEach((condition) => {
-        const propertyClasses = styleClasses[values[condition]];
+      Object.keys(propertyValue).forEach((condition) => {
+        const propertyClasses = styleClasses[propertyValue[condition]];
         if (!propertyClasses) {
+          handlePrimitive(propertyValue[condition], condition);
           return;
         }
-        classes.push(propertyClasses[condition]);
+        handlePrimitive(propertyClasses[condition], condition);
```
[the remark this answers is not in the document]
siriwatknp: It's working correct, see the test `should work with shorthands`.

siriwatknp (approved):
My last question is how this is going to be used in Material UI?

packages/pigment-css-react/src/generateAtomics.js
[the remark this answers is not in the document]
DiegoAndai: The test works correctly because it uses an array for the direction value, which is handled here: https://github.com/mui/pigment-css/blob/e3408002b3c155f915b5b26f76e343202dee254a/packages/pigment-css-react/src/generateAtomics.js#L52

The test fails if you provide `direction: { xs: 'row', sm: 'column }`, which should also be supported, and it's handled in the line we're discussing.

So besides changing this line we should add a test with all supported formats.

packages/pigment-css-react/src/Stack.d.ts
```diff
@@ -0,0 +1,20 @@
...
+import { PolymorphicComponent } from './Box';
+
+type CssProperty<T> = T | Array<T> | Partial<Record<Breakpoint, T>>;
+
+type StackBaseProps = {
+  display?: CssProperty<'flex' | 'inline-flex'>;
+  spacing?: CssProperty<number | string>;
+  direction?: CssProperty<CSS.StandardLonghandProperties['flexDirection']>;
+  justifyContent?: CssProperty<CSS.StandardLonghandProperties['justifyContent']>;
+  alignItems?: CssProperty<CSS.StandardLonghandProperties['alignItems']>;
```
DiegoAndai, line 13: We should remove `display`, `justifyContent`, and `alignItems`

apps/pigment-css-next-app/next.config.js
DiegoAndai: I'm not very familiar with the Pigment setup, why is this required in detail?

Is `displayName` no longer required?

DiegoAndai:
@siriwatknp 

> My last question is how this is going to be used in Material UI?

It will be reexported from Material UI. Pigment will be an optional peer dependency, and the import will fail if the user doesn't have pigment installed.

apps/pigment-css-next-app/next.config.js
siriwatknp: I'm not sure why @brijeshb42 removes the `displayName`, but for `overrideContext` it's a workaround for an error related to MUI System to make the app runnable (I will open a separate issue to discuss the options).

packages/pigment-css-react/src/Stack.d.ts
siriwatknp: Done.

packages/pigment-css-react/src/generateAtomics.js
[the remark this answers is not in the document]
siriwatknp: You are right. I pushed a fix with all the test cases that also cover `null`, `undefined` from the values.

packages/pigment-css-react/tests/generateAtomics.test.js
```diff
@@ -3,210 +3,39 @@ import { atomics } from '../src/generateAtomics';
...
-        sm: 'display-inline-flex-sm',
-        md: 'display-inline-flex-md',
-        lg: 'display-inline-flex-lg',
-        xl: 'display-inline-flex-xl',
-      },
-    },
     flexDirection: {
       column: {
         xs: 'flex-direction-column-xs',
-        $$default: 'flex-direction-column-xs',
```
brijeshb42: What are you assuming the default breakpoint in case someone passed this ?

```tsx
<Stack spacing={1} />
```

siriwatknp: Where is that `$$default` coming from? Isn't the `defaultCondition` enough to know the default breakpoint?

brijeshb42: Ok. It was a remnant of the first implementation of the `Box` component.

DiegoAndai (approved):

PR #136, opened 2024-06-12
[unplugin] Fix the path check for Windows

siriwatknp:
## Issue

On Windows, the received `id` is `...\@pigment-css\react\theme` but the unplugin only check for `/`. That's why the `generateThemeSource()` is not called which is result in `{}` for runtime theme and the build is error for components that try to use the theme.

**Before**: https://github.com/mui/material-ui/actions/runs/9475564524/job/26107055930?pr=42476

**After**: https://github.com/mui/material-ui/actions/runs/9476406034/job/26109235879?pr=42476


brijeshb42 (approved):
Thanks

PR #144, opened 2024-06-17
[react] Create Grid component

DiegoAndai:
Create Grid component following the atomics approach, same as https://github.com/mui/pigment-css/pull/118.

## Summary

This is the Pigment version of the Grid v2 component. It supports all the use cases the original supports but has some API differences:

### Sizing

Instead of using the breakpoints as props for size (`xs`, `sm`, `md`, ...), there's a `size` prop. Some conversions
- `xs={2}` turns into `size={2}`
- `xs={2} md={4}` turns into `size={{ xs: 2, md: 4 }}`
- `xs` turns into `size='grow'`
- `xs='auto'` turns into `size='auto'`
- `xs={2} sm md={4} lg='auto'` turns into `size={{ xs: 2, sm: 'grow', md: 4, lg: 'auto' }}`

### Offset
Instead of using the breakpoints as props for offset (`xsOffset`, `smOffset`, `mdOffset`, ...), there's a `offset` prop. Some conversions
- `xsOffset={2}` turns into `offset={2}`
- `xsOffset={2} mdOffset={4}` turns into `offset={{ xs: 2, md: 4 }}`
- `xsOffset='auto'` turns into `offset='auto'`
- `xsOffset={2} mdOffset={4} lgOffset='auto'` turns into `offset={{ xs: 2, md: 4, lg: 'auto' }}`

This was done as it works better with the no runtime approach. It might be possible to maintain the previous API, but I don't think it's worth it as it might introduce some tech debt, and this API is more consistent IMO. It is also easy to cover with a codemod.

### Unstable internal props.

The original Grid v2 uses a `unstable_level` prop for nesting grids. This uses a similar approach but instead of levels it takes advantage of CSS variable scoping. This is the only way I could find to achieve nested grids. The problem it solves is that when a Grid is both a container and an item, it has to read from and set the column count.

There might be alternatives to remove the use of these unstable internal props and check the component `muiName`, but I couldn't find a better way and I think we should move forward now. If someone has an idea for alternative implementations, I will test them. I don't think using `:has` would help.

## Play with the Grid

```bash
// project root
pnpm build

// pigment-css-vite-app
pnpm install
pnpm dev
```

The demo is at `http://localhost:5173/grid` and the code is at `apps/pigment-css-vite-app/src/pages/grid.tsx`

## Preview

I've copied all demos from the docs to the test apps.

https://github.com/brijeshb42/pigment-css/assets/16889233/42909b01-345c-4fc1-8374-2737ae7f7629


apps/pigment-css-next-app/src/app/grid/demo3.tsx
```diff
@@ -0,0 +1,43 @@
+'use client';
```
DiegoAndai, line 1: This is the only demo from the docs demos that is a client component, all others run as RSC 🚀 

The reason this has to be a client component is that it uses state to interactively show spacing changes.

packages/pigment-css-react/src/Grid.jsx
```diff
@@ -0,0 +1,311 @@
...
+    '--Grid-parent-column-spacing': ['--Grid-parent-column-spacing'],
+    '--Grid-parent-row-spacing': ['--Grid-parent-row-spacing'],
+    '--Grid-self-column-span': ['--Grid-self-column-span'],
+    '--Grid-self-width': ['--Grid-self-width'],
+    '--Grid-self-max-width': ['--Grid-self-max-width'],
+    '--Grid-self-flex': ['--Grid-self-flex'],
+    '--Grid-self-column-spacing': ['--Grid-self-column-spacing'],
+    '--Grid-self-row-spacing': ['--Grid-self-row-spacing'],
+    '--Grid-self-offset': ['--Grid-self-offset'],
+    '--Grid-self-margin-left': ['--Grid-self-margin-left'],
```
DiegoAndai, line 37: Yes, this is a lot of variables 😅 

But this is the only way I see we can support all the props being breakpoint aware: columns, spacing, size, offset, as well as nesting grids.

packages/pigment-css-react/src/Grid.jsx
```diff
@@ -0,0 +1,311 @@
...
+export const gridAtomics = generateAtomics(({ theme }) => ({
+  conditions: Object.keys(theme.breakpoints.values).reduce((acc, breakpoint) => {
+    acc[breakpoint] = `@media (min-width: ${theme.breakpoints.values[breakpoint]}${
+      theme.breakpoints.unit ?? 'px'
+    })`;
+    return acc;
+  }, {}),
+  defaultCondition: theme.breakpoints?.keys?.[0] ?? 'xs',
+  properties: {
+    flexDirection: ['column', 'column-reverse', 'row', 'row-reverse'],
```
DiegoAndai, line 26: This is the same as the Stack but I'm not sure we should abstract it, it might be too early. A little duplication is not that bad in my opinion.

packages/pigment-css-react/tests/generateAtomics.test.js
```diff
@@ -62,7 +65,7 @@ describe('generateAtomics', () => {
     ).to.deep.equal({
       className: 'gap--Stack-gap-lg gap--Stack-gap-xs',
       style: {
-        '--Stack-gap': 'calc(2 * 8px)',
+        '--Stack-gap-xs': 'calc(2 * 8px)',
```
DiegoAndai, line 68: I had to bring this back because there are some variable to variable mappings, for example `--Grid-self-width` that can be assigned to `--Grid-self-width-xs`, so we need different names from the variable that points to the breakpoint variables.

DiegoAndai:
As explained in the description, the APIs from Grid v2 to the Pigment Grid differ. It's easy to cover these differences with a codemod, which I think we should provide. The question would be should we implement this new API in the Grid v2?

siriwatknp:
For the sizing changes, is it an improvement or a mandatory change to make the Grid works?

packages/pigment-css-react/src/Grid.jsx
```diff
@@ -0,0 +1,311 @@
...
+}
+
+export const gridAtomics = generateAtomics(({ theme }) => ({
+  conditions: Object.keys(theme.breakpoints.values).reduce((acc, breakpoint) => {
+    acc[breakpoint] = `@media (min-width: ${theme.breakpoints.values[breakpoint]}${
+      theme.breakpoints.unit ?? 'px'
+    })`;
+    return acc;
+  }, {}),
+  defaultCondition: theme.breakpoints?.keys?.[0] ?? 'xs',
```
siriwatknp, line 24:
```suggestion
  defaultCondition: theme.breakpoints?.keys?.[0],
```
I think the default is not needed (I remove it as well from the Stack). Should keep the MUI System dependency as low as possible.

packages/pigment-css-react/src/Grid.jsx
siriwatknp: I agree with no abstraction. Each component should have independent config.

packages/pigment-css-react/src/Stack.jsx
```diff
@@ -23,7 +23,9 @@ const stackAtomics = generateAtomics(({ theme }) => {
       direction: ['flexDirection'],
       spacing: ['gap'],
     },
-    multiplier: Array.isArray(theme.vars?.spacing) ? theme.vars.spacing[0] : theme.vars?.spacing,
+    multipliers: {
+      gap: Array.isArray(theme.vars?.spacing) ? theme.vars.spacing[0] : theme.vars?.spacing,
+    },
```
siriwatknp, line 28: 👍

packages/pigment-css-react/src/Grid.jsx
```diff
@@ -0,0 +1,311 @@
...
+  if (container) {
+    GridAtomicsObj['--Grid-self-column-spacing'] = selfColumnSpacing;
+    GridAtomicsObj['--Grid-self-row-spacing'] = selfRowSpacing;
+  }
+
+  if (size) {
+    GridAtomicsObj['--Grid-self-column-span'] = size;
+    GridAtomicsObj['--Grid-self-width'] = size;
+    GridAtomicsObj['--Grid-self-max-width'] = size;
+    GridAtomicsObj['--Grid-self-flex'] = size;
```
siriwatknp, line 186: I think it is possible to support the existing `xs`, `sm`, … `xl` props without changing much code.

```suggestion
    GridAtomicsObj['--Grid-self-column-span'] = props;
    GridAtomicsObj['--Grid-self-width'] = props;
    GridAtomicsObj['--Grid-self-max-width'] = props;
    GridAtomicsObj['--Grid-self-flex'] = props;
```

packages/pigment-css-react/src/Grid.jsx
```diff
@@ -0,0 +1,311 @@
...
+  }
+  if (offset) {
+    GridAtomicsObj['--Grid-self-offset'] = offset;
+    GridAtomicsObj['--Grid-self-margin-left'] = offset;
```
siriwatknp, line 190:
```suggestion
    const offsetProps = {};
    Object.keys(props).forEach(prop => {
      if (prop.endsWith('Offset')) {
        offsetProps[prop.replace('Offset', '')] = props[prop]
      }  
    })
    GridAtomicsObj['--Grid-self-offset'] = offsetProps;
    GridAtomicsObj['--Grid-self-margin-left'] = offsetProps;
```

siriwatknp:
Please add the test for the component. You can follow this [pattern](https://github.com/mui/pigment-css/pull/146/files#diff-0e26807fa676c62d43763b45fe4e1dd7c815a5ecfe938fdbbf77422d5514f7f9).

packages/pigment-css-react/src/Grid.jsx
aarongarciah: Not a problem IMO 👍

aarongarciah:
@DiegoAndai about the API, some thoughts and questions:
- Ideally we should have the same API, no matter the styling solution.
  - _Side note:_ it's unfortunate we need to have different APIs based on the styling solution. These weird decisions is something we can improve in the future if we detach Material UI APIs from styling solutions as we discussed in other forums.
- I think the new API is way better and better aligned with how responsive values are handled using `sx` and other libraries across the industry.
- If we all consider the new API to be better, can we upgrade Grid v2 to use this new API and provide the codemod for the v6 upgrade?

packages/pigment-css-react/src/Grid.jsx
```diff
@@ -0,0 +1,311 @@
...
+    unstable_parent_row_spacing,
+    ...rest
+  },
+  ref,
+) {
+  const selfColumns = columns ?? unstable_parent_columns ?? 12;
+  const selfColumnSpacing = columnSpacing ?? spacing ?? unstable_parent_column_spacing ?? 0;
+  const selfRowSpacing = rowSpacing ?? spacing ?? unstable_parent_row_spacing ?? 0;
+
+  const GridAtomicsObj = {
```
brijeshb42, line 161:
```suggestion
  const gridAtomicsObj = {
```

packages/pigment-css-react/src/Grid.jsx
```diff
@@ -0,0 +1,311 @@
...
+    GridAtomicsObj['--Grid-self-flex'] = size;
+  }
+  if (offset) {
+    GridAtomicsObj['--Grid-self-offset'] = offset;
+    GridAtomicsObj['--Grid-self-margin-left'] = offset;
+  }
+
+  const ownerState = { container, size, offset };
+
+  const GridClasses = gridAtomics(GridAtomicsObj);
```
brijeshb42, line 195:
```suggestion
  const gridClasses = gridAtomics(gridAtomicsObj);
```

brijeshb42 (commented):
Left 2 minor nits. Rest all looks good. Nice work on implementing Grid. Next effort should be to have the similar API on runtime Grid as well for all layout components.

DiegoAndai:
> Please see [my suggestion](https://github.com/mui/pigment-css/pull/144/files#r1643657859), I'd prefer to keep the existing APIs as much as possible. I imagine that developers might switch back and forth between Emotion/Pigment CSS in v6.

@siriwatknp yes! your suggestion would work. I thought of implementing it that way, but I agree with @aarongarciah:

> I think the new API is way better and better aligned with how responsive values are handled using `sx` and other libraries across the industry.

In other words, I think having `size` and `offset` props is better. They're easier to understand, and they have the same usage as the `columns` or `spacing` props. I also think `'grow'` is easier to understand than the boolean `true` used in Grid v2.

So while I agree that the APIs should be the same if possible, I suggest keeping this API, deprecating the breakpoint props in Grid v2 (`xs`, `xsOffset`, `sm`, `smOffset`, ...) and implementing the `size` and `offset` props instead. We can have a codemod for it.

What do you think? Do you see any reason not to follow this approach?

packages/pigment-css-react/src/Grid.jsx
DiegoAndai: Updated

packages/pigment-css-react/src/Grid.jsx
DiegoAndai: Updated

brijeshb42 (approved):

packages/pigment-css-react/src/Grid.jsx
DiegoAndai: Updated

siriwatknp:
> So while I agree that the APIs should be the same if possible, I suggest keeping this API, deprecating (or removing) the breakpoint props in Grid v2 (xs, xsOffset, sm, smOffset, ...) and implementing the size and offset props instead. We can have a codemod for it.

To me, if the change in the API fixes issues I would consider going for it. However, it sounds like the change is based on our assumption that it is better. So if we compare "better API" with "more work + breaking change", I'd go with keeping the same API for now.

To the people who want to upgrade to v6, these breaking changes might not be better for them (since it is not fixing any issue). Even though we have codemod, it could not be enough. You will be surprised how fancy the codebase could look like when they open an issue that the codemod does not work 😅.

What do you think @mnajdova? Have you seen any issues regarding the existing API since older version @oliviertassinari?

DiegoAndai:
> However, it sounds like the change is based on our assumption that it is better.

I wouldn't say it's an assumption. It's objectively more consistent with the `sx`, `columns`, and `spacing` props. The naming is also objectively easier to understand: `xs` is a breakpoint, so why is it used as a name for a size prop?

> So if we compare "better API" with "more work + breaking change", I'd go with keeping the same API for now.

This is a good moment to improve the API: we are working on a new major, introducing an alternative no-runtime component, and Grid v2 is currently marked as unstable. This is the moment when we're focused on the layout components, I don't know if there will be another major soon that is focused on them.

brijeshb42:
Agree with Diego here. We can pitch this as a upcoming stable equivalent of Grid v2 and we have this time to do the API changes that we want.
Also agree with the new props.

DiegoAndai:
Hey @siriwatknp! I added the Grid tests setup. I wanted to add more tests to `Grid.test.js` but I wasn't able to import the Grid output from `fixtures/Grid.output` (like it's done for the [Hidden component for example](https://github.com/mui/pigment-css/blob/master/packages/pigment-css-react/tests/Hidden/Hidden.test.js#L26)). There's an issue with this:

```js
// fixtures/Grid.output.js

import _default from '@pigment-css/react';
// ...

const GridComponent = /*#__PURE__*/ _default('div')({
// ...
```


Which fails with: 

```
TypeError: (0 , _src.default) is not a function
```

I noticed the [Container fixture](https://github.com/mui/pigment-css/blob/master/packages/pigment-css-react/tests/Container/fixtures/Container.output.js#L5) also has this issue, so importing it doesn't work either. Do you know what might be going on?

cc: @brijeshb42

brijeshb42:
@DiegoAndai Modifying line 419 in `processors/styled.ts` to 

```js
const styledImportIdentifier = t.addNamedImport('styled', process.env.PACKAGE_NAME as string);
```

should fix your issue.

DiegoAndai:
Thanks @brijeshb42, it works now 👌🏼

DiegoAndai:
This is ready, the only thing left to decide is [the API discussion](https://github.com/mui/pigment-css/pull/144#issuecomment-2176136252).

siriwatknp:
> This is ready, the only thing left to decide is [the API discussion](https://github.com/mui/pigment-css/pull/144#issuecomment-2176136252).

To adopt the new API `size` and `offset`, these are extra works:
- update the Grid v2 to use the new props.
- create a codemod to transform into new props.

Let's vote, 👍 for yes and ❌ for no. cc @DiegoAndai @aarongarciah @brijeshb42 @mnajdova (@danilo-leal @zanivan tag both of you for user perspective)

apps/pigment-css-next-app/src/app/grid/demo3.tsx
mnajdova: AWESOME!!! ❤️

mnajdova:
> Let's vote, 👍 for Yes and 😕 for No

I want to justify my vote, as I haven't leave any comment so far. I am voting for the new API as it feels better, it's closer to the sx prop syntax that people are already familiar with. This is a major version, and the change is somewhat related to providing a support for new styling engine, considering we don't have almost any breaking change, it makes sense to include it, rather then adding more technical dept if we do want to change the API anyway in the next major.

siriwatknp:
@DiegoAndai based on https://github.com/mui/pigment-css/pull/144#issuecomment-2179809493, let's move forward with the new API. Can you update the Unstable_Grid in a separate PR?

DiegoAndai:
> @DiegoAndai based on https://github.com/mui/pigment-css/pull/144#issuecomment-2179809493, let's move forward with the new API. Can you update the Unstable_Grid in a separate PR?

Yes, I'll merge this today and open the PR to update the Unstable_Grid

siriwatknp:
@DiegoAndai I think `wrap` prop is missing. It's not a system prop (similar to `direction`). Can you fix it?

mnajdova:
Created PR https://github.com/mui/pigment-css/pull/159 for adding a wrap prop

PR #146, opened 2024-06-18
[react] Create Hidden component

siriwatknp:
## Summary

With a flip perspective, I am able to create the Hidden with the same API using the existing `generateAtomics` fn.

Instead of declaring `xs, … xl` as conditions, I use the prop name `xsUp, xsDown, xsOnly, … xlUp, xlDown, xlOnly` as conditions.

In the component, it reads the runtime theme to find the breakpoint keys and compose the `display` atomic object:

```js
// user code
<Hidden smUp>

// runtime
atomic({
  display: {
    smUp: 'none', // this will get the className for `smUp`
  }
})
```

## Test

See `packages/pigment-css-react/tests/generateAtomics.test.js`


packages/pigment-css-react/src/Hidden.jsx
```diff
@@ -0,0 +1,126 @@
...
+    <Component
+      ref={ref}
+      className={clsx(stackClasses.className, className)}
+      style={{ ...style, ...stackClasses.style }}
+      {...rest}
+    />
+  );
+});
+
+Hidden.propTypes = {
```
brijeshb42, line 59: You'll also need to wrap this with `process.env.NODE_ENV` check

packages/pigment-css-react/src/Hidden.jsx
```diff
@@ -0,0 +1,126 @@
...
+
+  return {
+    conditions,
+    properties: {
+      display: ['none'],
+    },
+  };
+});
+
+const Hidden = React.forwardRef(function Hidden(
```
brijeshb42, line 25: Can you also add test for `Hidden`'s runtime logic.

packages/pigment-css-react/src/Hidden.jsx
brijeshb42: I know that this is an edge case, but what will happen if multiple props are specified, like

`<Hidden smUp xlDown />`

siriwatknp: Good point, this works as expected.

siriwatknp: See `apps/pigment-css-next-app/src/app/hidden/page.tsx`

siriwatknp: > You'll also need to wrap this with process.env.NODE_ENV check

Done.

packages/pigment-css-react/src/Hidden.jsx
siriwatknp: Found the way, I transform the `Hidden` component before testing the runtime with `createRenderer`.

packages/pigment-css-react/src/index.ts
```diff
@@ -6,5 +6,4 @@ export { generateAtomics, atomics } from './generateAtomics';
 export { default as css } from './css';
 export { default as createUseThemeProps } from './createUseThemeProps';
 export { default as internal_createExtendSxProp } from './createExtendSxProp';
-export { default as Box } from './Box';
```
siriwatknp, line 9: @brijeshb42 I have to remove this, otherwise the test fail due to "Cannot find a module ./Box". I believe that Box should be imported from `@pigment-css/react/Box` right?

DiegoAndai (approved):

PR #149, opened 2024-06-19
[react] Remove sx transform logic and leverage Emotion unitless

### LINKED ISSUE #133: [RFC] shorthand properties & values should not be the default behavior
### What's the problem?

```js
import { styled } from '@pigment-css/react';

const A = styled.div({
  width: 1,
})
```

produces CSS `width: 100%`.

This is too magical and is hard to debug where it comes from.

### What are the requirements?

Pigment CSS should not have shorthand properties or values enabled by default. It should be configured explicitly based on the users.

### What are our options?

Pigment CSS needs to expose an API to configure the shorthand. I lean toward following Tailwind config so that users can switch between Tailwind and Pigment CSS.

### Proposed solution

Remove shorthand properties from the default behavior and let the users configure it explicitly. The API to configure is the next step to think about.

### Resources and benchmarks

_No response_

**Search keywords**:

siriwatknp:
closes #133

I found the comment about `unitless` and tried with `@emotion/unitless`. I don't see any problem. I am happy to revert the change if you think it should be a separate PR.

---


packages/pigment-css-react/src/utils/isUnitLess.ts
```diff
@@ -1,55 +1,4 @@
-// Copied over from https://github.com/emotion-js/emotion/blob/main/packages/unitless/src/index.js
-// since typings of @emotion/unitless are broken. PR to fix here - https://github.com/emotion-js/emotion/pull/3088
-const unitlessKeys: { [key: string]: 1 } = {
```
brijeshb42, line 3: This was added here since `@emotion/unitless`'s types were broken. Does it work now ?

brijeshb42:
Why is the lockfile changing so much ?

siriwatknp:
> Why is the lockfile changing so much ?

I open this PR before updating the package manager in package json. It's fixed now.

brijeshb42 (approved):

PR #152, opened 2024-06-19
[docs] Update readme to use next.config.mjs

mnajdova:
By default next-create-app@latest creates a `next.config.mjs`, so I've replaced the instructions that use `next.config.js` so use `next.config.mjs`.


brijeshb42:
Did full change here https://github.com/mui/pigment-css/pull/157 since mjs is now the default.

danilo-leal (approved):

PR #155, opened 2024-06-21
[react] Support transpiled library for `globalCss` and `keyframes`

siriwatknp:
Found this bug from https://github.com/mui/material-ui/pull/42640

## Root cause

The code of CssBaseline looks like this:

```js
const GlobalStyles = globalCss(({ theme }) => styles(theme));
```

However, after building Material UI as a package and Next.js import it, the code before passing to Pigment is:

```js
const GlobalStyles = globalCss(_c = ({
  theme
}) => styles(theme));
_c1 = GlobalStyles;
var c1;
```

Notice that the value passed to `globalCss` is a variable instead of a function. Pigment CSS thinks that it's not a function, that's why the CSS does not generated (with Emotion error):

```
Functions that are interpolated in css calls will be stringified.
If you want to have a css call based on props, create a function that returns a css call like this
let dynamicStyle = (props) => css`color: ${props.color}`
It can be called directly with props or interpolated in a styled call like this
let SomeComponent = styled('div')`${dynamicStyle}`
```

The fix is simple, we check the type after the evaluation and then pass the theme if it's a function.

---

### DIFF
--- a/packages/pigment-css-react/tests/globalCss/fixtures/globalCss-theme.output.css
+++ b/packages/pigment-css-react/tests/globalCss/fixtures/globalCss-theme.output.css
@@ -25,3 +25,6 @@
     color: red;
   }
 }
+html {
+  color: red;
+}
--- a/packages/pigment-css-react/src/processors/globalCss.ts
+++ b/packages/pigment-css-react/src/processors/globalCss.ts
@@ -134,18 +134,13 @@ export class GlobalCssProcessor extends BaseProcessor {
   }
 
   private handleCall([, callArg]: CallParam, values: ValueCache) {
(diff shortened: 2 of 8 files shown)


packages/pigment-css-react/tests/globalCss/fixtures/globalCss-theme.output.css
```diff
@@ -25,3 +25,6 @@
     color: red;
   }
 }
+html {
+  color: red;
+}
```
siriwatknp, line 30: Without the change, this will not be generated.

brijeshb42 (approved):

PR #159, opened 2024-06-24
[Grid] Add wrap prop

mnajdova:
Demo:

Screenshot 2024-06-24 at 10 59 09

### DIFF
--- a/packages/pigment-css-react/src/Grid.jsx
+++ b/packages/pigment-css-react/src/Grid.jsx
@@ -126,6 +126,12 @@ const GridComponent = styled('div')({
         marginLeft: 'var(--Grid-self-margin-left)',
       },
     },
+    ...['nowrap', 'wrap-reverse', 'wrap'].map((wrap) => ({
+      props: { wrap },
+      style: {
+        flexWrap: wrap,
+      },
+    })),
   ],
 });
 
@@ -150,6 +156,7 @@ const Grid = React.forwardRef(function Grid(
     unstable_parent_column_spacing,
     // eslint-disable-next-line react/prop-types
     unstable_parent_row_spacing,
+    wrap = 'wrap',
     ...rest
   },
   ref,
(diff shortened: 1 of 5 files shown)


packages/pigment-css-react/src/Grid.jsx
```diff
@@ -150,6 +156,7 @@ const Grid = React.forwardRef(function Grid(
     unstable_parent_column_spacing,
     // eslint-disable-next-line react/prop-types
     unstable_parent_row_spacing,
+    wrap = 'wrap',
```
siriwatknp, line 159:
```suggestion
    wrap,
```

There should be no default value for this, otherwise the non-container Grid will have `flexWrap` which is a different behavior compared to the previous version.

Grid container already contain the `flexWrap` by default (see the first variant).

siriwatknp: I was wrong.

packages/pigment-css-react/src/Grid.jsx
```diff
@@ -126,6 +126,12 @@ const GridComponent = styled('div')({
         marginLeft: 'var(--Grid-self-margin-left)',
       },
     },
+    ...['nowrap', 'wrap-reverse', 'wrap'].map((wrap) => ({
```
siriwatknp, line 129:
```suggestion
    ...['nowrap', 'wrap-reverse'].map((wrap) => ({
```
To match the previous behavior.

siriwatknp: `wrap` already assigned when container is true. or remove the `flexWrap` from the container variant.

siriwatknp (approved):
👍 Please check my comment before merging

packages/pigment-css-react/src/Grid.jsx
mnajdova: can this prop be applied to non-container grid? If no, I can update the props definition to include `container: true`

mnajdova: The previous implementation follows this ☝️ , I will update the code.

siriwatknp: > can this prop be applied to non-container grid

It should not. `flexWrap` only applied to flexbox container.

PR #191, opened 2024-08-05
[TypeScript] Fix `sx` types and missing `style` prop from layout components

siriwatknp:
Found this issue from https://github.com/mui/material-ui/pull/43065#pullrequestreview-2198658825

It's not possible to do this at the moment:

The `style` attribute is not supported in `PigmentGrid`

  image

### DIFF
--- a/packages/pigment-css-react/src/Stack.d.ts
+++ b/packages/pigment-css-react/src/Stack.d.ts
@@ -12,6 +12,8 @@ type StackBaseProps = {
   className?: string;
 };
 
-declare const Stack: PolymorphicComponent<StackBaseProps>;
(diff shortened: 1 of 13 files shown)


packages/pigment-css-react/src/Stack.d.ts
```diff
@@ -12,6 +12,8 @@ type StackBaseProps = {
   className?: string;
 };
 
-declare const Stack: PolymorphicComponent<StackBaseProps>;
```
brijeshb42, line 15: Can you add a typecheck test for the change if possible ?

siriwatknp: Done

PR #196, opened 2024-08-07
[core] Fix thrown error during eval phase

### LINKED ISSUE #193: [nextjs] Cannot use `sx` prop on layout components
### Steps to reproduce

Renders the Stack with sx prop:

```js
import Stack from "@mui/material-pigment-css/Stack";

<Stack sx={{}} />
```

### Current behavior

Got this error:

image

Without using `sx` prop, it works fine.

Also, vite has no issue, so I think this is nextjs-plugin specific

### Expected behavior

Should not throw errors

### Context

Found this issue from https://github.com/mui/material-ui/pull/43065#pullrequestreview-2198658825

### Your environment

<details>

```
  Don't forget to mention which browser you used.
  Output from `npx @mui/envinfo` goes here.
```
</details>

**Search keywords**: sx error

brijeshb42:
This PR changes the way we throw errors for our top-level exported functions. With the introduction of our build-time container components, we also started using the top-level APIs in these components. This in itself is ok since we don't want any restrictions. But this lead to an issue where an error was thrown, typically if trying to add sx prop on such container components (1 example).
This was because these container component files were also being evaluated during the eval phase of wyw and that phase also included calls to generateAtomics which always throws error. So with this change, we first detect whether the function is being called during the eval phase or not, by checking a special global variable called `__wyw_dynamic_import` which is set by WyW before evaluation happens. So now these functions don't throw error during the eval stage.

Fixes #193 

### DIFF
--- a/apps/pigment-css-next-app/src/app/grid/page.tsx
+++ b/apps/pigment-css-next-app/src/app/grid/page.tsx
@@ -13,7 +13,7 @@ const Item = styled.div`
 
 function GridDemo1() {
   return (
-    <Grid container spacing={2}>
+    <Grid container spacing={2} sx={{ color: '#070707' }}>
       <Grid size={8}>
         <Item>size=8</Item>
       </Grid>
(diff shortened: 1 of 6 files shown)


siriwatknp (approved):
Thanks for the explanation!

PR #200, opened 2024-08-08
[typescript] Fix `sx` types

siriwatknp:
## Issue

This [previous fix](https://github.com/mui/pigment-css/pull/191) causes Material UI build fails because the `SxProp` tries to get `theme` from an empty `ThemeArgs`.

The reason that Pigment CSS did not catch the error in [the PR](https://github.com/mui/pigment-css/pull/191) is because the project has augmented the theme types in `index.spec.ts` which causes the build to pass.

Will need to move all spec in a separate test environment so that it catches errors like this before merging.


brijeshb42 (approved):
Does this need a release?

siriwatknp:
> Does this need a release?

Yes.

PR #218, opened 2024-08-30
[plugin] Fix bug related to path separator on Windows

brijeshb42:
While debugging on Windows, I found that (atleast on Windows 11), the path separator was `/` instead of the expected `\` resulting in some files being skipped from being transformed from the `node_modules`.

So the fix first normalizes the path as per the platform and then checks for the `transformLibraries`'s presence in the normalized path.

Fixes - https://github.com/mui/material-ui/issues/43487

### DIFF
--- a/packages/pigment-css-vite-plugin/src/vite-plugin.ts
+++ b/packages/pigment-css-vite-plugin/src/vite-plugin.ts
@@ -132,17 +132,19 @@ export default function wywVitePlugin({
         .filter((m): m is ModuleNode => !!m);
     },
     async transform(code, url) {
-      const [id] = url.split('?', 1);
-
+      const [filePath] = url.split('?', 1);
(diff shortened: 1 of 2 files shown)


packages/pigment-css-vite-plugin/src/vite-plugin.ts
```diff
@@ -132,17 +132,19 @@ export default function wywVitePlugin({
         .filter((m): m is ModuleNode => !!m);
     },
     async transform(code, url) {
-      const [id] = url.split('?', 1);
-
+      const [filePath] = url.split('?', 1);
```
mnajdova, line 135: How does this arg differs from the one in the unplugin (url vs filePath)?

brijeshb42: No difference. Updated to use same names.

PR #219, opened 2024-08-30
[system] Add recent change in function name to patch

brijeshb42:
The internal export has been renamed from `internal_processStyles` to `internal_serializeStyles` in a recent [PR](https://github.com/mui/material-ui/pull/43412) resulting in build failures. This change fixes that. Ideally, this should be moved to the core repo so that similar changes in future do not require a fix in this repo.

- [x] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/pigment-css/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).

PR #220, opened 2024-08-30
fix: styled-engine new internals

romgrk:
Update to match the changes in https://github.com/mui/material-ui/pull/43412


apps/pigment-css-next-app/next.config.js
romgrk, line 1: Shouldn't the next.js config be abstracted in a plugin like the vite one is?

romgrk:
Duplicate of https://github.com/mui/pigment-css/pull/219

packages/pigment-css-react/src/index.ts
```diff
@@ -8,3 +8,4 @@ export { default as createUseThemeProps } from './createUseThemeProps';
 export { default as internal_createExtendSxProp } from './createExtendSxProp';
 export { default as useTheme } from './useTheme';
 export { default as globalCss } from './globalCss';
+export { default as internal_styledEngineMockup } from './styledEngineMockup';
```
brijeshb42, line 11: This should not be part of the public api (base module export) but can come from a path like `@pigment-css/react/internal`.

romgrk: I followed the convention for `internal_createExtendSxProp` right above. I'll move it to a new internal folder.

packages/pigment-css-react/src/internal/index.ts
brijeshb42, line 1: You should also add this path to the `exports` in package.json

romgrk: I've tried it but it wasn't enough, not sure how the different `(tsconfig|tsup).*` files interact, any advice to setup the build properly?

romgrk: Nevermind, I think I got it.

apps/pigment-css-next-app/next.config.js
```diff
@@ -118,13 +120,7 @@ const pigmentOptions = {
       ...context,
       require: (id) => {
         if (id === '@mui/styled-engine' || id === '@mui/styled-engine-sc') {
-          return {
```
brijeshb42, line 121: This is not needed here. I'll clean this up separately since it's not a package.

packages/pigment-css-react/src/internal/index.ts
brijeshb42, line 1: One last change needed is an entry it `.gitignore` and include `internal` in `package.json`'s `files` key.

romgrk:
Should be good now.

brijeshb42 (approved):

romgrk:
@brijeshb42 Can the netlify CI be ignored? And if so, could you merge this PR? I don't have write access to this repo.

I'm also not sure how pigment-css releases work, when will this be released? If it's released by next week then I'll look into finalizing the PR on the core.

romgrk:
Ah I think I need this to be published to merge the core one, or the tests won't pass. Much appreciated if you can release it tomorrow.

brijeshb42:
@romgrk Can you check your changes in core with [this](https://ci.codesandbox.io/status/mui/pigment-css/pr/224/builds/537949) CI package to make sure the changes work before I actually publish ?

romgrk:
How would I go about setting that up? I see the http CSB package but iiuc that only works on CSB.

I've checked by linking my local pigment repo to my local material repo. With the old branch if fails due to the internal API mismatch, and with the new branch if fails due to unrelated errors, so I'm assuming it's picking up the right thing.

brijeshb42:
You can use the csb urls in the top level package.json's `resolutions` key and then run the build.
Make sure to add all the 4 packages in the key.
If youve already tested then 👍

brijeshb42:
Published `v0.0.22`. Please check your changes with this version.

Issue #199, opened 2024-08-07
Component styles are overridden when rendered inside Next.js layout


corydeppen:
### Search keywords

nextjs vite plugin layout components override

### Latest version

- [X] I have tested the latest version

### Steps to reproduce

Review the styles being applied to the nav links in the mobile appbar of both Next.js and Vite versions of the example app. Observe how the styles are incorrect when on the home page of the Next app but are correct on the about page.

### Current behavior

I've run into an issue that appears when using the Next.js plugin but doesn't exist when using the Vite plugin. As you'll see in the [example app](https://stackblitz.com/edit/github-xv6opg?file=src%2Fapp%2F(public)%2Fcomponents%2Fsite-app-bar.tsx), the nav links in the appbar drawer on mobile don't have the correct alignment or padding specified by the `ListItemButton` styles. However, if you flip to the About page and inspect the links in the drawer, you'll see the `ListItemButton` styles are being applied as expected.

From what I can tell, it looks like the Next.js plugin is generating styles at a route file level (e.g. `page.tsx`, `layout.tsx`) instead of at a component/module level like the Vite plugin appears to do. In the Next app, `(public)/page.tsx` has another `Button` on the page so the base button styles from `page.tsx` take precedence over the styles being generated for `layout.tsx` that renders the `SiteAppBar` component containing the links. In the [Vite example app](https://stackblitz.com/edit/tanstack-router-pxcte6?file=src%2Froutes%2F-components%2Fsite-app-bar.tsx), though, the styles are being generated separately for the `site-app-bar.tsx` module and take precedence so the `ListItemButton` alignment and padding are applied as expected.

### Expected behavior

Styles for components rendered within Next.js layout components should not be overridden by those in page components.

### Your environment

<details>
  <summary><code>npx @mui/envinfo</code></summary>

```
System:
  OS: Linux 5.0 undefined
Binaries:
  Node: 18.20.3 - /usr/local/bin/node
  npm: 10.2.3 - /usr/local/bin/npm
  pnpm: 8.15.6 - /usr/local/bin/pnpm
Browsers:
  Chrome: Not Found
npmPackages:
  @emotion/react:  11.13.0 
  @emotion/styled:  11.13.0 
  @mui/core-downloads-tracker:  6.0.0-dev.240424162023-9968b4889d 
  @mui/icons-material: 6.0.0-beta.4 => 6.0.0-beta.4 
  @mui/material: 6.0.0-beta.4 => 6.0.0-beta.4 
[...]

PR #226, opened 2024-09-10
[next] Wrap all generated CSS with a cascade layer (WIP)

Janpot:
One way to fix https://github.com/mui/pigment-css/issues/199

* Updated [stackblitz](https://stackblitz.com/edit/github-xv6opg-fggtsn?file=package.json) for https://github.com/mui/pigment-css/issues/199
* Updated [codesandbox](https://codesandbox.io/p/github/wh5938316/mui-v6-pigment-css/csb-ymg78v/draft/floral-dream) for https://github.com/mui/material-ui/issues/43529.

Notes:

* Does this have unwanted side-effects?
* TODO: sourcemaps. Can we get away with them shifting down 1 line for the sake of a quick fix? => let's run with it for now.
* Does this belong in `wyw-in-js`?

Closes https://github.com/mui/material-ui/pull/43705

### DIFF
--- a/packages/pigment-css-unplugin/src/index.ts
+++ b/packages/pigment-css-unplugin/src/index.ts
@@ -284,6 +284,15 @@ export const plugin = createUnplugin<PigmentOptions, true>((options) => {
         }
 
         let { cssText } = result;
+
+        const slug = slugify(cssText);
+
+        cssText = `
+          @layer _${slug} {
+            ${cssText}
+          } 
+        `;
+
         if (isNext && !outputCss) {
           return {
             code: result.code,
@@ -324,7 +333,7 @@ export const plugin = createUnplugin<PigmentOptions, true>((options) => {
         }
 
         const rootPath = process.cwd();
-        const slug = slugify(cssText);
+
         const cssFilename = path
           .normalize(`${id.replace(/\.[jt]sx?$/, '')}-${slug}.pigment.css`)
           .replace(/\\/g, path.posix.sep);
--- a/packages/pigment-css-unplugin/src/index.ts
+++ b/packages/pigment-css-unplugin/src/index.ts
@@ -26,7 +26,6 @@ import {
 } from '@pigment-css/react/utils';
 import type { ResolvePluginInstance } from 'webpack';
 import { styledEngineMockup } from '@pigment-css/react/internal';
-
 import { handleUrlReplacement, type AsyncResolver } from './utils';
 
 type NextMeta = {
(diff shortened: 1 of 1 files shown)


DiegoAndai:
> Does this have unwanted side-effects?

We have visual regression tests running with Pigment on the Material UI repo, we can create a draft PR pointing to these builds to check if anything unexpected is happening there. It probably won't cover all edge cases but it would be helpful.

> Can we get away with them shifting down 1 line for the sake of a quick fix?

DiegoAndai:
> Does this have unwanted side-effects?

We have visual regression tests running with Pigment on the Material UI repo, we can create a draft PR pointing to these builds to check if anything unexpected is happening there. It probably won't cover all edge cases but it would be helpful.

> Can we get away with them shifting down 1 line for the sake of a quick fix?

Why do you consider this to be a quick fix?

If it is a quick fix, can users implement it on their side? We could propose it as a workaround until we agree on a definitive solution.

> Does this belong in wyw-in-js?

If it is, let's try to fix it on our side anyway. We can open an issue on their repo and refactor it later, but let's not block a fix on our side.

Janpot:
> Why do you consider this to be a quick fix?

Because it's just rewriting the css without considering the source maps. This means that the links in devtools will be off by one line. It may be quite a bit more involved to write this in a way that sourcemaps are transformed as well. It all comes down to how much effort we want to pour down into fixing this.

> If it is a quick fix, can users implement it on their side?

I don't see how.

brijeshb42:
I am not sure how fool proof the fix is and if wouldn't cascade into some other bug.
Let me check this change with some calls to globalCss api.

Janpot:
What do you mean with "globalCss api"? 

I've tested by adding a `./global.css` overriding some properties on [Mui classes](https://mui.com/material-ui/api/button/#classes). That still works since unlayered styles are more specific than layered styles.

packages/pigment-css-unplugin/src/index.ts
```diff
@@ -284,6 +283,19 @@ export const plugin = createUnplugin<PigmentOptions, true>((options) => {
         }
 
         let { cssText } = result;
+
+        const slug = slugify(cssText);
```
brijeshb42, line 287: Can you move this at the end of the `if (isNext)` check at line https://github.com/mui/pigment-css/blob/master/packages/pigment-css-unplugin/src/index.ts#L301

since we mainly want to do this for next.js and not webpack.

brijeshb42:
> Does this have unwanted side-effects?
We'll have to see. It fixes the current bug.

> TODO: sourcemaps. Can we get away with them shifting down 1 line for the sake of a quick fix? => let's run with it for now.

Sourcemaps are working without any issues since the linkage is through the generated class names which we are not touching.

> Does this belong in wyw-in-js?

Not really since wyw does not have any nextjs specific solution. If it had `@wyw/nextjs`, then yes.

brijeshb42 (approved):
Thanks for the fix.

PR #233, opened 2024-09-12
[Grid] ignore RSC error

siriwatknp:
closes https://github.com/mui/material-ui/issues/43635

This is a temporary fix, which might not cover an edge case if the consumer use client component as a child of Grid container.

However, this fix should work for normal use cases because a Grid container should always have Grid item.

### DIFF
--- a/packages/pigment-css-react/src/Grid.jsx
+++ b/packages/pigment-css-react/src/Grid.jsx
@@ -8,10 +8,16 @@ import { generateAtomics } from './generateAtomics';
 import styled from './styled';
 
 function isGridComponent(element) {
-  // For server components `muiName` is avaialble in element.type._payload.value.muiName
-  // relevant info - https://github.com/facebook/react/blob/2807d781a08db8e9873687fccc25c0f12b4fb3d4/packages/react/src/ReactLazy.js#L45
-  // eslint-disable-next-line no-underscore-dangle
-  return element.type.muiName === 'Grid' || element.type?._payload?.value?.muiName === 'Grid';
+  try {
+    // For server components `muiName` is avaialble in element.type._payload.value.muiName
+    // relevant info - https://github.com/facebook/react/blob/2807d781a08db8e9873687fccc25c0f12b4fb3d4/packages/react/src/ReactLazy.js#L45
+    // eslint-disable-next-line no-underscore-dangle
+    return element.type.muiName === 'Grid' || element.type?._payload?.value?.muiName === 'Grid';
+  } catch (error) {
+    // Covers for the case in which the Grid is a server component and the child is a client component
+    // https://github.com/mui/material-ui/issues/43635
+    return false;
+  }
 }
 
 const gridAtomics = generateAtomics(({ theme }) => {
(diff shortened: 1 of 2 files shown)


DiegoAndai:
Thanks for taking this @siriwatknp.

The solution doesn't work for this case: https://mui.com/material-ui/react-grid2/#inheriting-columns. For that use case to work, a grid item must forward its parent column and spacing values, which won't work if we check for the container prop.

What do you think about something like this?:

```js
function isGridComponent(element) {
  try {
    // For server components `muiName` is avaialble in element.type._payload.value.muiName
    // relevant info - https://github.com/facebook/react/blob/2807d781a08db8e9873687fccc25c0f12b4fb3d4/packages/react/src/ReactLazy.js#L45
    // eslint-disable-next-line no-underscore-dangle
    return element.type.muiName === 'Grid' || element.type?._payload?.value?.muiName === 'Grid';
  } catch {
    // Covers for the case in which the parent is a server component and the child is a client component
    // https://github.com/mui/material-ui/issues/43635
    return false
  };
}
```

Would this work? It would fail silently, but maybe we can add a callout on the docs about this case.

We should also catch only this specific error and not others.

siriwatknp:
> The solution doesn't work for this case: https://mui.com/material-ui/react-grid2/#inheriting-columns.

I don't understand why it would not work. The logic I added is to scoped the check to Grid container. In the demo you provide, all the grid still receive the `unstable_*` props.

Argos does not show any difference, so it should be the same right?

siriwatknp:
> What do you think about something like this?:

You proposal works! I will go with your proposal.

packages/pigment-css-react/src/Grid.jsx
```diff
@@ -8,10 +8,16 @@ import { generateAtomics } from './generateAtomics';
...
-  // For server components `muiName` is avaialble in element.type._payload.value.muiName
-  // relevant info - https://github.com/facebook/react/blob/2807d781a08db8e9873687fccc25c0f12b4fb3d4/packages/react/src/ReactLazy.js#L45
-  // eslint-disable-next-line no-underscore-dangle
-  return element.type.muiName === 'Grid' || element.type?._payload?.value?.muiName === 'Grid';
+  try {
+    // For server components `muiName` is avaialble in element.type._payload.value.muiName
+    // relevant info - https://github.com/facebook/react/blob/2807d781a08db8e9873687fccc25c0f12b4fb3d4/packages/react/src/ReactLazy.js#L45
+    // eslint-disable-next-line no-underscore-dangle
+    return element.type.muiName === 'Grid' || element.type?._payload?.value?.muiName === 'Grid';
+  } catch (error) {
```
DiegoAndai, line 16: Can we check for the specific error so we don't catch unintended ones?

```js
if (error.message.startsWith("Cannot access default.muiName on the server.") {
  // Covers for the case in which the Grid is a server component and the child is a client component
    // https://github.com/mui/material-ui/issues/43635
    return false;
} else {
  throw error
}
```

siriwatknp: Are you sure, the message check is quite fragile. Given the size of the function, eventhough it catches, I think it's still better than checking message.

cc @brijeshb42

DiegoAndai (approved):

brijeshb42 (approved):
Just speculating. Is there some way we can also give this control to the user so they can customize this if the error criteria changes in the future?

PR #234, opened 2024-09-12
[react] Fix css extraction for transformed styled tagged-template call

brijeshb42:
1. Fixes https://github.com/mui/material-ui/issues/43619
2. Also adds default preprocessor in `unplugin` package.

The main issue here is that `next.js` through `swc` transpiles files in `node_modules` which also includes tagged template literals, converting them to a normal `styled()()` with somewhat weird arguments.
This transformation was not handled previously resulting in some missing css from the final output.

### DIFF
--- a/packages/pigment-css-react/src/processors/styled.ts
+++ b/packages/pigment-css-react/src/processors/styled.ts
@@ -14,6 +14,7 @@ import {
   IOptions as IBaseOptions,
 } from '@wyw-in-js/processor-utils';
 import {
+  type FunctionValue,
   ValueType,
   type ConstValue,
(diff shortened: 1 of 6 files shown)


brijeshb42:
@Janpot I see some side-effects of `@layer` fix that we did while testing the example in the original bug. Button's styles are not being applied.

Without `@layer` :

Screenshot 2024-09-12 at 5 52 08 PM

With `@layer` :

Screenshot 2024-09-12 at 5 53 15 PM

Ripple related fixes are working fine.

siriwatknp (approved):
Great job on this! @brijeshb42

PR #240, opened 2024-09-17
[react] Wrap individual css output into their own cascade layer

### LINKED ISSUE #199: Component styles are overridden when rendered inside Next.js layout

brijeshb42:
Also, update the base style output to include pigment cascade layer order.

This change wraps the CSS output into its own layers and then the base css style output determines the precedence of the layers.

Lets assume this to be kind of an RFC PR to discuss the usage and impact of Cascade layers.

One immediate downside I see is increase in generated CSS size since each individual css output is wrapped in a layer.
Endusers might need to include a PostCSS plugin to merge layers,media queries and classNames.

This fixes #199 

### DIFF
--- a/packages/pigment-css-unplugin/src/index.ts
+++ b/packages/pigment-css-unplugin/src/index.ts
@@ -302,17 +302,17 @@ export const plugin = createUnplugin<PigmentOptions, true>((options) => {
           }
 
           // Valid names must start with a underscore or letter.
-          const layerName = `_${slug}`;
+          // const layerName = `_${slug}`;
 
-          // Fix for https://github.com/mui/pigment-css/issues/199
-          // Bring each file in its own layer so that the order is maintained between css modules
-          // shared between layout.tsx and page.tsx.
-          // TODO: Do this in a way that keeps the source map correct
-          cssText = `
-            @layer pigment.${layerName} {
-              ${cssText}
-            }
-          `;
+          // // Fix for https://github.com/mui/pigment-css/issues/199
+          // // Bring each file in its own layer so that the order is maintained between css modules
+          // // shared between layout.tsx and page.tsx.
+          // // TODO: Do this in a way that keeps the source map correct
+          // cssText = `
+          //   @layer pigment.${layerName} {
+          //     ${cssText}
+          //   }
+          // `;
         }
 
         if (sourceMap && result.cssSourceMapText) {
--- a/packages/pigment-css-react/src/processors/styled.ts
+++ b/packages/pigment-css-react/src/processors/styled.ts
@@ -249,7 +249,7 @@ export class StyledProcessor extends BaseProcessor {
 
     const baseClass = this.getClassName();
     this.baseClasses.push(baseClass);
-    this.collectedStyles.push([baseClass, cssText, null]);
+    this.collectedStyles.push([baseClass, `@layer pigment.base {${cssText}}`, null]);
     const variantsAccumulator: VariantData[] = [];
     this.processOverrides(values, variantsAccumulator);
     variantsAccumulator.forEach((variant) => {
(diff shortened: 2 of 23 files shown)


o-alexandrov:
If this issue is purely in Next.js, why do you make this change for Vite also?

brijeshb42:
The original issue is for next.js but the solution here is generic that we were anyways planning to implement. This would have helped with better css precedence for base and variant styles.
This as a side-effect also fixes the layout issue.

packages/pigment-css-unplugin/src/index.ts
```diff
@@ -302,17 +302,17 @@ export const plugin = createUnplugin<PigmentOptions, true>((options) => {
...
-              ${cssText}
-            }
-          `;
+          // const layerName = `_${slug}`;
+
+          // // Fix for https://github.com/mui/pigment-css/issues/199
+          // // Bring each file in its own layer so that the order is maintained between css modules
+          // // shared between layout.tsx and page.tsx.
+          // // TODO: Do this in a way that keeps the source map correct
+          // cssText = `
```
mnajdova, line 311: I don't think we are solving the same issue. We discussed with briefly with @Janpot, namely the layering for the base, overrides and variants styles should be there, but we should not remove this, because the conflict we had in the previous issue comes from defining two base styles (as we use the styled() call on a styled component):

```
// these are both styles that would be added in the base layer, no?
const Paper = styled(div)({});
const AppBar = styled(Paper)({});
```

PR #243, opened 2024-09-18
feat: adapt to pre-processed styles

romgrk:
Adapt to pre-processed styles to match https://github.com/mui/material-ui/pull/43412

### DIFF
--- a/packages/pigment-css-react/src/processors/styled.ts
+++ b/packages/pigment-css-react/src/processors/styled.ts
@@ -620,7 +620,7 @@ export class StyledProcessor extends BaseProcessor {
       );
     }
     if (styleObj.isProcessed) {
-      styleObj = styleObj.style
+      styleObj = styleObj.style;
     } else {
       delete styleObj.variants;
     }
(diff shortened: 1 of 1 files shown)


packages/pigment-css-react/src/processors/styled.ts
```diff
@@ -619,7 +619,11 @@ export class StyledProcessor extends BaseProcessor {
         })),
       );
     }
-    delete styleObj.variants;
+    if (styleObj.isProcessed) {
```
brijeshb42, line 622: Should this check instead be for the presence of the key `isProcessed` instead of checking for true?

romgrk: This feels good to me, any reason for checking that way instead of this one?

brijeshb42: checking for key's presence makes sure that this object is coming from the new processor regardless of the value.

romgrk: The value can only be `true`, it makes more sense to check for `true` than for the key presence.

PR #254, opened 2024-09-29
[core] Increase the minimum Node.js version support to 14.0.0

oliviertassinari:
Same minimum version increase as in https://github.com/mui/material-ui/pull/42920, from v12.0.0 to v14.0.0. No reason for the version to be different.


oliviertassinari:
@brijeshb42 what do you think about this one?

brijeshb42:
Makes sense to me.

brijeshb42 (approved):

PR #266, opened 2024-10-09
[docs] Readme tweaks

aarongarciah:
Some README updates:
- Update table of contents.
- Add missing indentation to a couple of list items.
- Switch list items for headings in a list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -892,7 +892,7 @@ On the other hand, Pigment CSS extracts CSS at build time and replaces the Java
 
 Here are some common patterns and how to achieve them with Pigment CSS:
 
-1. **Fixed set of styles**
+#### 1. Fixed set of styles
 
 In Emotion or styled-components, you can use props to create styles conditionally:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -345,43 +351,43 @@ There are two ways to achieve this (both involve using a CSS variable):
 
 1. Declare a CSS variable directly in the styles and set its value using inline styles:
 
-```jsx
-const Heading = styled('h1')({
-  color: 'var(--color)',
-});
+   ```jsx
```
DiegoAndai, line 354: Is the code blocks indentation expected? Other code blocks here aren't.

aarongarciah: Yes, it's expected. These code blocks are nested inside list items, so they should be indented. Other code blocks are not indented because they're not inside list items (unless I missed any).

|before|after|
|---|---|
|Screenshot 2024-10-09 at 20 47 03|Screenshot 2024-10-09 at 20 44 50|

DiegoAndai (approved):

PR #271, opened 2024-10-11
[core] Fix test_types_next job

Janpot:
Noticed this was failing on master.


mnajdova (approved):
The next dependency was already updatd, I kept the changes in the circleci/config.yml

PR #286, opened 2024-10-24
[react] Fix scoping issue with sx dynamic value transform

brijeshb42:
For this sample component -

```js
function App(props) {
  return (
    <SliderRail
      {...props}
      sx={[
        props.variant === 'secondary'
          ? { color: props.isRed ? 'red' : 'blue' }
          : { backgroundColor: 'blue', color: 'white' },
      ]}
    />
  );
}
```

In the current sx transform, we were transforming the dynamic part to an arrow function with the same arguments that it was using internally and then later in the `sx.ts` based wyw transform, we were just replacing the arrow function with just its body (to be same as user code). This was being done to play well with wyw were it throws an error if it tries to evaluate an object that references local variables.

Current transform -

```js
sx={[
  props.variant === 'secondary'
    ? { color: (props) => props.isRed ? 'red' : 'blue' }
    : { backgroundColor: 'blue', color: 'white' },
]}
```

This did not play well with minifiers where the variables in transformed function would be renamed like -

```js
 (_props) => _props.isRed ? 'red' : 'blue'
```

So later when we were trying to revert back to user code, it'll just be

```js
{ color: _props.isRed ? 'red' : 'blue' }
```

resulting in an issue trying to use a variable that does not exist. This was more apparent when using webpack with it's terser based uglifier.

This PR fixes the issue by serializing the AST as a string and later on deserializing it and generating the code from the AST.

Partially fixes [#43597](https://github.com/mui/material-ui/issues/43597)

### DIFF
--- a/packages/pigment-css-react/src/processors/sx.ts
+++ b/packages/pigment-css-react/src/processors/sx.ts
@@ -99,6 +99,19 @@ export class SxProcessor extends BaseProcessor {
         ([variableId, expression, isUnitLess]) => {
           switch (expression.type) {
             case 'ArrowFunctionExpression': {
+              const fnBody = expression.body as Expression;
+              // try to deserialize AST if possible
+              if (
+                fnBody.type === 'StringLiteral' &&
+                (fnBody.value[0] === '{' || fnBody.value[0] === '[')
+              ) {
+                try {
+                  const body = JSON.parse(fnBody.value);
+                  return t.objectProperty(t.stringLiteral(variableId), body);
+                } catch (ex) {
+                  // proceed as usual
+                }
+              }
               return t.objectProperty(
                 t.stringLiteral(variableId),
                 t.arrayExpression([expression.body as Expression, t.booleanLiteral(isUnitLess)]),
--- a/packages/pigment-css-react/src/processors/sx.ts
+++ b/packages/pigment-css-react/src/processors/sx.ts
@@ -101,10 +101,7 @@ export class SxProcessor extends BaseProcessor {
             case 'ArrowFunctionExpression': {
               const fnBody = expression.body as Expression;
               // try to deserialize AST if possible
-              if (
-                fnBody.type === 'StringLiteral' &&
-                (fnBody.value[0] === '{' || fnBody.value[0] === '[')
-              ) {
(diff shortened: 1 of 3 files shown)


siriwatknp (commented):
Looks good. Is it possible to add a test case for this?

brijeshb42:
Not specifically for this. The motivation here was also that existing tests do not break regardless of the internal changes.
So in a way, we already had tests.
Actual test might require an elaborate setup with terser of some other uglifier.

siriwatknp:
> Not specifically for this. The motivation here was also that existing tests do not break regardless of the internal changes. So in a way, we already had tests. Actual test might require an elaborate setup with terser of some other uglifier.

I see.

packages/pigment-css-react/src/utils/sxObjectExtractor.ts
```diff
@@ -93,10 +101,17 @@ function traverseObjectExpression(
...
-          const arrowFn = arrowFunctionExpression(
-            localIdentifiers.map((i) => i.node),
-            cloneNode(value.node),
-          );
+          // eslint-disable-next-line no-nested-ternary
+          const cssKey = key.isIdentifier()
+            ? key.node.name
+            : key.isStringLiteral()
+              ? key.node.value
+              : '';
```
siriwatknp, line 109: better to remove ternary or remove this lint config.

siriwatknp (approved):
👍 Please check my comment.

packages/pigment-css-react/src/processors/sx.ts
```diff
@@ -99,6 +99,19 @@ export class SxProcessor extends BaseProcessor {
         ([variableId, expression, isUnitLess]) => {
           switch (expression.type) {
             case 'ArrowFunctionExpression': {
+              const fnBody = expression.body as Expression;
+              // try to deserialize AST if possible
+              if (
```
mnajdova, line 104: Do we even need this condition here? We anyway have the code wrapped in try & catch.

brijeshb42: Removed other checks but `StringLiteral` check is needed for type narrowing.

PR #294, opened 2024-11-01
core: Bump package versions and fix related TS errors

brijeshb42:
There was one breaking change in Emotion. Some common types had been moved to `@emotion/serialize`.

There is also a small change in how we do conversion for !important. Currently, we do it by replacing ! with __IMP__ and then reconverting back.
In this PR, I have changed that to just encode and decode to and from base64. So we don't need special handling for `!` character.

I encountered these errors while setting up the Pigment CSS docs workspace.


apps/pigment-css-next-app/package.json
```diff
@@ -10,18 +10,18 @@
...
     "react-dom": "^18.3.1",
-    "next": "latest"
+    "next": "15.0.2"
   },
   "devDependencies": {
     "@pigment-css/nextjs-plugin": "workspace:^",
-    "@types/node": "^20.5.7",
-    "@types/react": "^18.3.3",
-    "@types/react-dom": "^18.2.19",
+    "@types/node": "^18.19.63",
```
siriwatknp, line 22: This's a downgrade. Is this correct?

packages/pigment-css-nextjs-plugin/src/virtual-css-loader.js
```diff
@@ -1,6 +1,6 @@
 export const loader = function virtualFileLoader() {
   const callback = this.async();
   const resourceQuery = this.resourceQuery.slice(1);
-  const { source } = JSON.parse(decodeURIComponent(resourceQuery));
-  return callback(null, source.replaceAll('__IMP__', '!important'));
+  const { source } = JSON.parse(decodeURIComponent(decodeURI(atob(resourceQuery))));
```
mnajdova, line 4: 👍

packages/pigment-css-react/src/processors/css.ts
```diff
@@ -9,7 +9,7 @@ import type {
 } from '@wyw-in-js/processor-utils';
 import type { Replacements, Rules } from '@wyw-in-js/shared';
 import { ValueType } from '@wyw-in-js/shared';
-import type { CSSInterpolation } from '@emotion/css';
+import type { CSSInterpolation } from '@emotion/serialize';
```
mnajdova, line 12: How did this breaking change land in an minor 😕

apps/pigment-css-next-app/package.json
brijeshb42: It is. v20 never actually was used since at the root, in `resolutions`, it was still v18. I have kept it at 18 since we are currently using node v18 to build in CI.

apps/pigment-css-next-app/next.config.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-env node */
 // eslint-ignore-next-line import/no-unresolved
 const { withPigment } = require('@pigment-css/nextjs-plugin');
-const { experimental_extendTheme: extendTheme } = require('@mui/material/styles');
+const { extendTheme } = require('@mui/material/styles');
```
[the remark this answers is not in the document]
brijeshb42: Sure. I'll update it in the actual docs PR.

apps/pigment-css-vite-app/package.json
```diff
@@ -10,31 +10,31 @@
...
   "dependencies": {
     "@pigment-css/react": "workspace:^",
-    "@mui/utils": "next",
-    "@mui/base": "next",
-    "@mui/lab": "next",
-    "@mui/material": "next",
-    "@mui/system": "next",
-    "@mui/icons-material": "next",
-    "clsx": "^2.1.0",
+    "@mui/utils": "6.1.6",
```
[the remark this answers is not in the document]
brijeshb42: IMO, it makes sense for `examples`. But for local apps, I'd prefer to manually update the versions.

packages/pigment-css-react/src/processors/css.ts
brijeshb42: Yeah. I was surprised as well.

PR #304, opened 2024-11-06
[docs] More docs infra changes

brijeshb42:
* Separate text content and it's wrapper's styles
* Use env variables for constants
* Use docs infra config from @mui/monorepo
* Correctly link md or mdx files to Github
* Hide version selector if it has only one version
* Simplify and enable bottom pagination style
* Setup code highlighting of mdx code content with built-in shiki theme

Screenshot 2024-11-06 at 4 51 49 PM


docs/next.config.ts
```diff
@@ -1,20 +1,38 @@
...
 
-const DATA_DIR = path.join(process.cwd(), 'data');
+const currentDirectory = url.fileURLToPath(new URL('.', import.meta.url));
+const DATA_DIR = path.join(currentDirectory, 'data');
 
 const nextConfig: NextConfig = {
   trailingSlash: false,
   env: {
     DATA_DIR,
+    CURRENT_VERSION: rootPackage.version,
```
mnajdova, line 19: Where is this used?

docs/src/components/AppBar.tsx
```diff
@@ -45,7 +45,9 @@ export function AppBar() {
         <IconLinkButton useNextLink href="/" label="Pigment CSS" size={3}>
           <PigmentIcon />
         </IconLinkButton>
-        <DocsVersionSelector currentVersion={currentVersion} versions={supportedVersions} />
+        {supportedVersions.length > 1 && (
```
mnajdova, line 48: Instead of hiding we can have a span to show the current version.

docs/next.config.ts
brijeshb42: To be used in version dropdown (or span).

docs/src/components/AppBar.tsx
brijeshb42: Will update in a followup.

brijeshb42:
> It would be ideal to have a use-cases for all changes in order to be able to test them in more details.

Makes sense. I was testing locally with stubs. Going forward, most of the PRs for docs will be about the actual content. So we'll get a better idea then.

PR #310, opened 2024-11-09
[nextjs] Fix path resolution

### LINKED ISSUE #171: withPigment causes tsconfig compiler path option to not work
### Steps to reproduce

1. With a Next.js 13+ project, configure withPigment in next.config.js
2. In the typescript tsconfig.json file, specify the compiler path option of 
`"paths":` {"src/*": ["./src/*"]},`
3. Add an import such as: 
`import {getStateName from 'src/common/utils'`
4 When compiling the next app, the error` EvalError: Cannot find module 'src/common/utils'` will be experienced
5. Eliminating withPigment will eliminate the error (but styles will be broken).  Changing the import to a relative path will correct the error

### Current behavior

Compiler error of `EvalError: Cannot find module `

### Expected behavior

No compiler error

### Context

Use a typescript remapped path

### Your environment

<details>

```
  Don't forget to mention which browser you used.
  Output from `npx @mui/envinfo` goes here.
```
</details>

**Search keywords**: withPigment typescript path Next.js

brijeshb42:
This reverts the previous "fix" and uses the newly available `loaderContext` from unplugin to do path resolution. Previous solution was "hacky" in a sense since at that time, Unplugin did not expose loader's context. So we had to do things manually which did not work reliably. Now that we have "loaderContext" data, it's straighforward to just call `loaderContext.resolve()` with the path and this works as expected.

Tested on our docs and user submitted repro repository.

Fixes #171 

### DIFF
--- a/packages/pigment-css-unplugin/src/index.ts
+++ b/packages/pigment-css-unplugin/src/index.ts
@@ -8,7 +8,7 @@ import {
   transform,
   createFileReporter,
 } from '@wyw-in-js/transform';
-import { asyncResolveFallback, slugify } from '@wyw-in-js/shared';
+import { slugify } from '@wyw-in-js/shared';
 import {
   UnpluginFactoryOutput,
   WebpackPluginInstance,
@@ -25,7 +25,6 @@ import {
   type PluginCustomOptions,
 } from '@pigment-css/react/utils';
 import { styledEngineMockup } from '@pigment-css/react/internal';
-
 import { handleUrlReplacement, type AsyncResolver } from './utils';
 
 type NextMeta = {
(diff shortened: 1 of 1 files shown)


packages/pigment-css-unplugin/src/index.ts
```diff
@@ -344,7 +330,7 @@ export const plugin = createUnplugin<PigmentOptions, true>((options) => {
           };
         }
 
-        const rootPath = process.cwd();
+        const rootPath = projectPath;
```
siriwatknp, line 333: Can you remove this `rootPath` and use `projectPath` directly.

valleywood:
@brijeshb42 How does the release process look like (when will there be a 0.0.27 patch release)?

brijeshb42:
Published new version `0.0.27`.

valleywood:
Thanks @brijeshb42 can confirm all my aliases working again with version `0.0.27` 👍

PR #341, opened 2024-12-04
feat: Implement @pigment-css/theme package

brijeshb42:
At this point, it is just a helper library for extending TS types of the user theme.

This package is a bunch of TS helpers to easily allow users to override the `Theme` type which will be used in `@pigment-css/core` and `@pigment-css/react` packages.

This is the 1st package in v1 series. Next PRs on top will actually use this package.

### DIFF
--- a/packages/pigment-css-theme/tests/theme.spec.ts
+++ b/packages/pigment-css-theme/tests/theme.spec.ts
@@ -1,0 +1,14 @@
+import { t } from '../src';
+
+declare module '../src' {
+  interface Theme {
+    palette: {
+      main: string;
+    };
+  }
+}
+
+t('$palette.main');
+
+// @ts-expect-error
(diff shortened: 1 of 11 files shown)


packages/pigment-css-theme/tests/theme.spec.ts
```diff
@@ -0,0 +1,14 @@
...
+    palette: {
+      main: string;
+    };
+  }
+}
+
+t('$palette.main');
+
+// @ts-expect-error
+t('$palette.secondary');
```
mnajdova, line 14: 😍

mnajdova (approved):
We'll need to be careful with these complex types and make sure we don't get trapped into the TS check performance issues we had in Material UI. I am happy to try and see how it goes :) 

Let's merge it.

PR #350, opened 2024-12-17
fix: use the correct serializer based on the direction value

brijeshb42:
Curently, the assumption was that the rtl css would only be needed as an addon.
Fixed this to always use an explicit direction serializer based on the direction.
Also, the Vite plugin was not passing on the direction information to pre-processor. This has been fixed.

Fixes https://github.com/mui/material-ui/issues/44647

Tested on user's repo with a local build of Pigment.

Screenshot 2024-12-17 at 5 32 20 PM

### DIFF
--- a/packages/pigment-css-react/package.json
+++ b/packages/pigment-css-react/package.json
@@ -73,7 +73,7 @@
     "react": "^18.3.1"
   },
   "peerDependencies": {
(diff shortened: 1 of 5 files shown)


mnajdova (approved):
Logic & tests looks good 👌

PR #364, opened 2025-01-09
fix: spread style object before deleting variants

brijeshb42:
This was mutating the original theme object resulting in different codes for server and client components.

Fixes: https://github.com/mui/material-ui/issues/44675


siriwatknp (approved):
👍

PR #371, opened 2025-01-21
[pigment-core] Create new UI framework independent package

brijeshb42:
### `@pigment-css/core`

1. Supports `css()` and `keyframes()` api
2. Add preprocessor to generate keyframes' correct output
3. `processors/css.ts` encompasses all the core logic that'll be inherited by `styled` in `@pigment-css/react` as well.
4. All generated css are part of different layers `pigment.base`, `pigment.variants` and `pigment.compoundvariants` with preference in the same order.
5. In each `css()` call, users can define how the class name is to be generated instead of this being a top-level pigment config, ie,

```js
// cls's value will be random but deterministic across builds
const cls = css`
  color: red;
`;
```

```js
// cls's value will always be "logo"
const cls = css({className: 'logo'})`
  color: red;
`;
```

#### TODOS

- [ ] `globalCss` implementation


mnajdova (commented):
API surface and fixtures output looks good. Tagging @Janpot or @siriwatknp for a second review.

mnajdova (approved):

packages/pigment-css-core/README.md
```diff
@@ -0,0 +1,29 @@
...
+Pigment CSS supports Next.js and Vite with support for more bundlers in the future.
+
+### Why choose Pigment CSS
+
+Thanks to recent advancements in CSS (like CSS variables and `color-mix()`), "traditional" CSS-in-JS solutions that process styles at runtime are no longer required for unlocking features like color transformations and theme variables which are necessary for maintaining a sophisticated design system.
+
+Pigment CSS addresses the needs of the modern React developer by providing a zero-runtime CSS-in-JS styling solution as a successor to tools like Emotion and styled-components.
+
+Compared to its predecessors, Pigment CSS offers improved DX and runtime performance (though at the cost of increased build time) while also being compatible with React Server Components.
+Pigment CSS is built on top of [WyW-in-JS](https://wyw-in-js.dev/), enabling to provide the smoothest possible experience for Material UI users when migrating from Emotion in v5 to Pigment CSS in v6.
```
Janpot, line 16: This sentence feels a bit outdated given that pigment won't be a requirement for Material UI, and v6 will be legacy soon.

packages/pigment-css-core/src/css.js
```diff
@@ -0,0 +1,8 @@
+import { generateErrorMessage } from './utils';
+
+export default function css() {
+  console.error(generateErrorMessage('css'));
```
Janpot, line 4: Wouldn't it make more sense to throw here? Users may miss the error message and be puzzled about why no styles appear even though it seemingly is working. Besides that, wouldn't this produce a ton of duplicate messages in the console?

packages/pigment-css-core/src/keyframes.js
```diff
@@ -0,0 +1,10 @@
+import { generateErrorMessage } from './utils';
+
+/**
+ * `__wyw_dynamic_import` is a special global var that is set during the evaluation phase by wyw.
+ * So during eval phase, it can happen that some code is calling the runtime function.
+ * We do not want to throw error in that case as we want the evaluation to happen.
+ */
+export default function keyframes() {
```
Janpot, line 8: Is there a reason not to author those files in typescript? Aren't they currently being built by `tsup`?

Janpot: Wouldn't that work if the function returns a `string`?

packages/pigment-css-core/exports/css.js
```diff
@@ -0,0 +1,5 @@
+Object.defineProperty(exports, '__esModule', {
```
Janpot, line 1: Are these to serve node10 resolution?

packages/pigment-css-core/README.md
Janpot: Docs are to be written separately I suppose?

siriwatknp (commented):
Overall looks good. I think adding comment to test files and may be link to the feature/public documentation will make it even better.

Are you planning to add a documentation in this PR?

packages/pigment-css-core/src/base.ts
```diff
@@ -0,0 +1,51 @@
+import type * as CSS from 'csstype';
+import { Theme, ThemeKey } from '@pigment-css/theme';
+
+export type CSSProperties = CSS.PropertiesFallback<number | string> & {
+  [index: `--${string}`]: string;
+  [index: `$${string}`]: string;
```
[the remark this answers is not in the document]
Janpot: Is that a pigment specific thing?

Janpot (approved):
It's a bit hard to get a comprehensive view on the features proposed without documentation. My feeling is that there would rather be a bit too much API than too little. But we can judge that when we review the docs and retroactively remove or change.

It is to say: let's treat the docs PR as the one where we review API and make changes to it.

PR #383, opened 2025-02-20
Allow `false` in the sx TypeScript types

bartlangelaan:
- [X] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/pigment-css/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).

The following snippet is from the documentation:

```
<div
  sx={[
    { color: 'red' },
    selected && { fontWeight: 'bold' },
    disabled ? (theme) => ({ opacity: theme.state.disabledOpacity }) : { opacity: 1 },
  ]}
/>
```

However, this generates a TypeScript error that `boolean` is not allowed in the sx prop.

This PR allows for `false` values to be in the SxProp array, like it is allowed in `@mui/system`.

### DIFF
--- a/packages/pigment-css-react/tests/sx.spec.ts
+++ b/packages/pigment-css-react/tests/sx.spec.ts
@@ -11,6 +11,7 @@ const sx3: SxProp = [{ color: 'red' }, { backgroundColor: 'blue', color: 'white'
 const test = true;
 const sx4: SxProp = [
   test ? { color: 'red' } : { backgroundColor: 'blue', color: 'white' },
+  false,
   (theme) => ({
     border: `1px solid ${theme.palette.primary.main}`,
   }),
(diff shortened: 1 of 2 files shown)


packages/pigment-css-react/tests/sx.spec.ts
```diff
@@ -11,6 +11,7 @@ const sx3: SxProp = [{ color: 'red' }, { backgroundColor: 'blue', color: 'white'
 const test = true;
 const sx4: SxProp = [
   test ? { color: 'red' } : { backgroundColor: 'blue', color: 'white' },
+  false,
```
siriwatknp, line 14: Can you add another test to verify this?

```ts
let selected: boolean | undefined;
const sx4: SxProp = [
  selected && { color: 'red' },
]
```

PR #390, opened 2025-03-14
[docs] All the docs actual content

brijeshb42:

### DIFF
--- a/docs/src/content/guides/how-it-works.mdx
+++ b/docs/src/content/guides/how-it-works.mdx
@@ -1,0 +1,120 @@
+# How Pigment CSS works
+
+Pigment CSS is a zero-runtime CSS-in-JS library. This means it does not have access to the end user's browser runtime which would be necessary to generate and insert authored CSS at runtime. Instead, it does all its processing at build time to pre-generate the CSS which then becomes part of the output bundle. This is why it cannot be consumed on its own. You must install the Pigment CSS plugin that corresponds to the framework you're using and configure your bundler accordingly (See the [integrations](/overview/integrations) page for more information).
+
+## Processor
+
+Pigment CSS uses the [WyW-in-JS](https://wyw-in-js.dev/) library that also powers [Linaria](https://linaria.dev/). It features a [processor](https://wyw-in-js.dev/how-to/custom-tagged-template#creating-a-processor) which makes it possible to create custom logic that's triggered by the presence of different imports from the library. The processor looks through the source code for `styled()`, `css()`, and other function calls and extracts the arguments to be evaluated. These values are then handed back to Pigment CSS for additional parsing and evaluation.
+
+For example, here's a simple usage of the `css()` function:
+
+```js
+// app.js
+import { css } from '@pigment-css/react-new';
+
+const testClass = css(({ theme }) => ({
+  padding: 0,
+  position: 'relative',
+  ...theme.typography.body1,
+}));
+```
+
+This snippet uses the `css()` call to generate and assign a class name string to the `testClass` variable.
+
+The corresponding (minimal) bundler configuration for this to work with the theme might look like:
+
+```ts title="bundler.config.ts"
+const customTheme = {
+  typography: {
+    body1: {
+      fontFamily: '"Roboto", "Helvetica", "Arial", sans-serif',
+      fontWeight: 400,
+      fontSize: 16,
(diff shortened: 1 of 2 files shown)


docs/src/content/features/composition.mdx
```diff
@@ -0,0 +1,254 @@
...
+function App() {
+  return (
+    <div className={`${bgBlue} ${bgRed}`}>
+      <p>Hello, world!</p>
+    </div>
+  );
+}
+```
+
+In the above example, the color of the text will be `blue` in both the cases because the final generated css will have the styles in the same order -
```
Janpot, line 44: Gotta love `stylex` for fixing this. It's so counterintuitive that just swapping the position of two global variables affects the behavior of where they are applied. I mean, imagine the following two snippets print a different result.

```tsx
const a = 5;
const b = 7;
console.log(a / b);
// ---
const b = 7;
const a = 5;
console.log(a / b);
```

Janpot: > Both have different outputs with same order of variable declaration.

Exactly, in javascript, the order at call site is important, not at declaration site.

> OTOH, Pigment's output is the same as when you'd write CSS modules.

Indeed, that's what's counterintuitive, I'm writing javascript, but I have to keep CSS semantics in mind. The real danger comes when utilities are defined in other files and composed. Now you don't have a module-local record of the order of declaration. 

> Besides, the selling point of stylex is the same as tailwind (when used with something like tailwind-merge), which is atomic css.

Sure this is done under the hood with atomic CSS and a lightweight runtime, but I don't think that matters. It's not at all its selling point to me, I don't care how they solve this problem, I care that the semantics match the environment I'm writing in, which is javascript, not css.

docs/src/content/overview/integrations.mdx
```diff
@@ -0,0 +1,147 @@
...
+};
+
+const pigmentConfig: PigmentCSSConfig = {
+  // Pigment CSS configuration options here
+};
+
+export default withPigment(nextConfig, pigmentConfig);
+```
+
+In the top-level `layout.tsx` file, add an import to the `@pigment-css/react-new/styles.css` file at the top.
```
Janpot, line 70: pages router docs seem to be missing?

docs/src/content/features/styling.mdx
```diff
@@ -0,0 +1,465 @@
...
+
+After you have [configured your bundler](/overview/integrations), you can start styling your application using Pigment CSS.
+
+## css
+
+This is the most basic API in Pigment CSS that allows you to style your application using CSS in a framework agnostic way.
+
+You can use either the tagged-template syntax of the JS object syntax to style your application. You can also mix and match the two as per the needs.
+
+### Tagged-template syntax
```
Janpot, line 18: I'm likely going to use only one of the syntaxes exclusively in my application. Seeing the other one in the dosc feels a bit like clutter. Would it make sense to turn this into a selector like the "tailwind css"/"css modules" one from the base ui docs?

docs/src/content/guides/how-it-works.mdx
```diff
@@ -0,0 +1,120 @@
...
+  position: 'relative',
+  ...theme.typography.body1,
+});
+
+module.exports = {
+  _exp1,
+};
+```
+
+The above code is then evaluated through the use of the `node:module` [module](https://nodejs.org/docs/v20.11.1/api/modules.html#module) which returns the evaluated values of the parameters.
```
Janpot, line 60: Do you mean https://nodejs.org/docs/v20.11.1/api/vm.html ?

Janpot: Also, I'd explain that the result of this step is that it obtains a static CSS fragment that is associated with the `css` call.

In the next step I'd say that now it generates a classname it can use in place of the `css()` call on js side, and use to wrap the styles with on CSS side

docs/src/content/features/composition.mdx
o-alexandrov: > The real danger comes when utilities are defined in other files and composed. Now you don't have a module-local record of the order of declaration. You can't even reliably influence it as it may already have been imported somewhere else in a different order.

We've hit this problem w/ pigment a few times; just want to highlight this problem is real in <5 frontend devs projects, and should be a lot worse when many devs in a project attempt to reuse each others' styles from many different files

PR #406, opened 2025-06-06
fix: separator issue for Windows

Cary123:
- [ ] I have followed (at least) the [PR section of the contributing guide](https://github.com/mui/pigment-css/blob/HEAD/CONTRIBUTING.md#sending-a-pull-request).

### DIFF
--- a/packages/pigment-css-vite-plugin/src/index.ts
+++ b/packages/pigment-css-vite-plugin/src/index.ts
@@ -66,10 +66,10 @@ export function pigment(options: PigmentOptions) {
       name: 'pigment-css-theme-injection-plugin',
       enforce: 'pre',
       resolveId(source) {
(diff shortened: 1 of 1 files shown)


packages/pigment-css-vite-plugin/src/index.ts
```diff
@@ -66,10 +66,10 @@ export function pigment(options: PigmentOptions) {
       name: 'pigment-css-theme-injection-plugin',
       enforce: 'pre',
       resolveId(source) {
-        if (finalTransformLibraries.some((lib) => source.includes(`${lib}/styles.css`))) {
+        if (allLibs.some((lib) => source.includes(`${lib}/styles.css`))) {
```
brijeshb42, line 69: Could you move this check to a util function, use that here and add some test that were previously failing.

